PR #61, opened 2021-08-08
Include `require` statements into README

gjtorikian:
It took me way too long to figure out the gem's name. 😅


alexrudall:
Thank you @gjtorikian!

PR #125, opened 2022-12-14
Adding Org Header

maks112v:
I started getting issues today on a project that org header was missing. I added it in the same way as access token it can be passed as an env variable or directly on class init.

## All Submissions:


alexrudall:
Thanks a lot @maks112v!

PR #142, opened 2022-12-30
Fix Files upload API

chaadow:
## All Submissions:

Hi,

The Files#upload API expects a form-data body instead of a JSON one. by adding this simple check this makes it work for all APIs.
I've looked at the [file upload VCR cassette](https://github.com/alexrudall/ruby-openai/blob/main/spec/fixtures/cassettes/files_upload.yml) to make sure it wasn't JSON, and it was using form-data as well.

### DIFF
--- a/lib/ruby/openai/client.rb
+++ b/lib/ruby/openai/client.rb
@@ -50,7 +50,7 @@ module OpenAI
       HTTParty.post(
         uri(path: path),
         headers: headers,
-        body: parameters
+        body: path == '/files' ? parameters : parameters.to_json
       )
     end
 
--- a/lib/ruby/openai/client.rb
+++ b/lib/ruby/openai/client.rb
@@ -50,7 +50,7 @@ module OpenAI
       HTTParty.post(
         uri(path: path),
         headers: headers,
-        body: parameters.to_json
+        body: parameters
       )
     end
 
@@ -67,7 +67,7 @@ module OpenAI
 
     private_class_method def self.headers
       {
-        "Content-Type" => "application/json",
+        # "Content-Type" => "application/json",
         "Authorization" => "Bearer #{Ruby::OpenAI.configuration.access_token}",
         "OpenAI-Organization" => Ruby::OpenAI.configuration.organization_id
       }
--- a/lib/ruby/openai/client.rb
+++ b/lib/ruby/openai/client.rb
@@ -67,7 +67,7 @@ module OpenAI
 
     private_class_method def self.headers
       {
-        # "Content-Type" => "application/json",
+        "Content-Type" => "application/json",
         "Authorization" => "Bearer #{Ruby::OpenAI.configuration.access_token}",
         "OpenAI-Organization" => Ruby::OpenAI.configuration.organization_id
       }


lib/ruby/openai/client.rb
```diff
@@ -50,7 +50,7 @@ def self.post(path:, parameters: nil)
       HTTParty.post(
         uri(path: path),
         headers: headers,
```
chaadow, line 52: also the content type is wrong, but open-ai didn't care about it. maybe add an optional headers param. rename the `self.headers` to `self.defaut_headers`

then do `default_headers.merge(headers)`. 

also I feel it's not very scalable the way I did the conditional. the proper way is that each business class  calls itself `parameters.to_json`... Make it explicit

and in here, we only do : `body: parameters` without the .to_json call

chaadow: it would be a bit verbose, unfortunately, but cleaner IMO

petergoldstein:
So I have an alternative suggestion, which lines up with the refactoring I'm doing to support multi-tenant.

Looking through the OpenAI API documentation, there are really two types of POSTs:

1. Posts using type multipart/form-data - `/files/upload`, `/fine-tunes/#{id}/cancel`, `/images/edits`, `/images/variations`
2. Posts using type application/json - `/fine-tunes`, `/images/generations`, `/completions`, `/edits`, `/embeddings`, `/moderations`

The fact that these are all POSTs should not constrain us to use the same method for both.  So if we introduce a multipart_post and a json_post and update the calls appropriately, this all works out:

```
    def json_post(path:, parameters: nil)
      HTTParty.post(
        uri(path: path),
        headers: headers,
        body: parameters&.to_json
      )
    end

    def multipart_post(path:, parameters: parameters)
      HTTParty.post(
        uri(path: path),
        headers: headers(content_type: "multipart/form-data"),
        body: parameters
      )
    end

    def headers(content_type: "application/json")
      {
        "Content-Type" => content_type,
        "Authorization" => "Bearer #{access_token}",
        "OpenAI-Organization" => organization_id
      }
    end

```

I'm going to incorporate the above change into my larger rewrite branch.  (Note this already has the multi-tenant changes)

chaadow:
> So I have an alternative suggestion, which lines up with the refactoring I'm doing to support multi-tenant.
> 
> Looking through the OpenAI API documentation, there are really two types of POSTs:
> 
> 1. Posts using type multipart/form-data - `/files/upload`, `/fine-tunes/#{id}/cancel`, `/images/edits`, `/images/variations`
> 2. Posts using type application/json - `/fine-tunes`, `/images/generations`, `/completions`, `/edits`, `/embeddings`, `/moderations`
> 
> The fact that these are all POSTs should not constrain us to use the same method for both. So if we introduce a multipart_post and a json_post and update the calls appropriately, this all works out:
> 
> ```
>     def json_post(path:, parameters: nil)
>       HTTParty.post(
>         uri(path: path),
>         headers: headers,
>         body: parameters&.to_json
>       )
>     end
> 
>     def multipart_post(path:, parameters: nil)
>       HTTParty.post(
>         uri(path: path),
>         headers: headers(content_type: "multipart/form-data"),
>         body: parameters
>       )
>     end
> 
>     def headers(content_type: "application/json")
>       {
>         "Content-Type" => content_type,
>         "Authorization" => "Bearer #{access_token}",
>         "OpenAI-Organization" => organization_id
>       }
>     end
> ```
> 
> I'm going to incorporate the above change into my larger rewrite branch. (Note this already has the multi-tenant changes)

@petergoldstein Great suggestion. Since you're on the subject of rewriting, I would suggest being able to override/merge headers in each business class ( Finetunes/ Files etc.). in addition to. the multipart_post json_post.

```ruby

def multipart_post(path:, parameters: nil, custom_headers: {})
       HTTParty.post(
         uri(path: path),
         headers: headers(content_type: "multipart/form-data").merge(custom_headers),
         body: parameters
       )
```

petergoldstein:
@chaadow I guess I don't understand the point of custom headers here.  What use case are you trying to address?

The OpenAI API uses a very limited set of HTTP headers.  If it starts using one or more additional headers, that's easy to add.

Moreover, adding the ability to override the headers set by the gem (Content-Type, authorization related headers) would allow breakage that shouldn't be possible.  That's bad.

To me custom headers don't sound like a valuable add.  I'd like to understand what you have in mind.

alexrudall:
@chaadow @petergoldstein thanks so much for your contributions here, I've gone with @petergoldstein's https://github.com/alexrudall/ruby-openai/pull/153

Specs now pass and can be run without VCR so they actually hit the API. This will cost about 2 cents. You'll need to add your `OPENAI_ACCESS_TOKEN=` in `.env`.

```
  NO_VCR=true bundle exec rspec
```

Issue #149, opened 2023-01-03
Allow clients to have different OpenAI configurations


petergoldstein:
**Is your feature request related to a problem? Please describe.**

I want to build a platform such that I can use multiple OpenAI credentials and configurations.  For example, I may have a SaaS that uses models from different organizations.  Or I may want to tier my users, such that one tier of users gets access to OpenAI at a higher usage threshold.  In these cases I want a single application to be able to define a Client object with a given configuration for a per-user context.  Currently that's not possible with this library.

**Describe the solution you'd like**

I thought that the global configuration would be a default, that could be overridden on creation of a `OpenAI::Client` object.  Instead, based on the code [here](https://github.com/alexrudall/ruby-openai/blob/0519baa062cb97037ffb8fa6c7bae8ac42f59ee6/lib/ruby/openai/client.rb#L5), invocation of a client overwrites the global configuration.  

Converting the code to a global default, overriddable on a per-client basis, would be fairly straightforward for everything except the `organization_id` (since you might have a non-null default, which could be overridden on a per-client basis, potentially with nil).

Leaving this problem aside, this change could be accomplished with the following steps:

1. The initializer for `OpenAI::Client` would be rewritten to set per-instance values for `access_token`, `organization_id`, and `api_version` based on arguments and default to those stored in `Ruby::OpenAI::configuration` if nil (or a NullObject for `organization_id`)
2. There'd be a general conversion of the `OpenAI::Client` class methods to instance methods, potentially leaving the existing public class methods in place as ones that use the 'default' client (one that uses the global configuration)

I recognize that this would be a pretty significant change, but as-is the gem can't support any real multi-tenant use case. My thoughts would be that this might make sense as a v4 change.

**Describe alternatives you've considered**

None really.  This is how most libraries of this nature (including [Dalli](https://github.com/petergoldstein/dalli), which I maintain) handle multi-tenancy with default configurations. The organization_id makes it somewhat more ambiguous, but could probably be solved pretty easily with a NullObject pattern in the `OpenAI::Client` initializer.

**Additional context**

I'd be happy to put up a PR if there's interest. 

It might also make sense to look at supporting connection pools as a further step towards supporting usage in a multi-tenant SaaS.

alexrudall:
[...]

PR #150, opened 2023-01-03
Support multi-tenant use of the gem

petergoldstein:
Addresses #149 

This is a quick and dirty conversion to multi-tenancy for the gem.  That said, both Rubocop and specs pass and I expect it works in practice.

Some changes I'd suggest on top of this commit:

1. Make the Files, Finetunes, Images, and Models classes non-instantiable except via the Client object.  All access to these APIs should go through a Client object.  This would dramatically simplify the initializers for these classes
2. Potentially add a `default_client` class method to OpenAI::Client, which would produce a memoized OpenAI::Client object that is initialized with the global parameters
3. Consider swapping out Httparty for connection pools, to minimize latency under production conditions
4. Maybe exclude the organization header from the header set if the value is nil?  I don't think we need to send in that case.
5. Allow the global configuration to have a missing access key, and shift that error to the Client upon initialization.

I don't think these changes make the gem any more difficult to use.  The only change from a user perspective is that setting the API credentials on a single client doesn't change it for the whole app.  So if you don't set the global credentials, you will need to invoke every Client instance with credentials.

## All Submissions:

### DIFF
--- a/lib/ruby/openai/client.rb
+++ b/lib/ruby/openai/client.rb
@@ -2,51 +2,60 @@ module OpenAI
   class Client
     URI_BASE = "https://api.openai.com/".freeze
 
-    def initialize(access_token: nil, organization_id: nil)
-      Ruby::OpenAI.configuration.access_token = access_token if access_token
-      Ruby::OpenAI.configuration.organization_id = organization_id if organization_id
+    NULL_ORGANIZATION_ID = Object.new.freeze
+
+    attr_reader :access_token, :organization_id, :api_version
+
+    def initialize(access_token: nil, organization_id: NULL_ORGANIZATION_ID, api_version: nil)
+      @access_token = access_token || Ruby::OpenAI.configuration.access_token
+      @organization_id = if organization_id == NULL_ORGANIZATION_ID
+                           Ruby::OpenAI.configuration.access_token
+                         else
+                           organization_id
+                         end
+      @api_version = api_version || Ruby::OpenAI.configuration.api_version
     end
 
     def completions(parameters: {})
-      OpenAI::Client.post(path: "/completions", parameters: parameters)
+      post(path: "/completions", parameters: parameters)
     end
 
     def edits(parameters: {})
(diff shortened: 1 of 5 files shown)


alexrudall:
@petergoldstein thanks a lot for this! It looks great.

If I run some of the specs without VCR (so actually hitting the API, this is a new feature I added last week) 

```
  NO_VCR=true bundle exec rspec spec/ruby/openai/client/completions/codex_spec.rb
```

I get failures with this error

```
 "message": "No such organization: <OPENAI_ACCESS_TOKEN>.",
```

Do you know why this is?

petergoldstein:
@alexrudall Yes, there was a bug in the code (I had fixed it in the other PR but not here) where the wrong default value was being used when the organization_id is not set.  I just updated it here.  There are specs in the other PR that catch this, but since I moved the code around a bit it's not a trivial pull-back.

bf4:
@petergoldstein @alexrudall looks good, though needs a rebase against main now.  Any blockers to this?

bf4:
@alexrudall any way to revive this PR? The Twitter gem did the same thing many years ago-- switched from being a global client to an instantiatable client.  I think it's probably a good pattern and already has a contribution.

bf4:
I'm happy to revive this if there's still interest. I think it's good work.

bf4:
This has been open since January with maintainer interest. I recommend closing.

MarioRuiz:
right now we don't have any option to have client instances with different hosts and/or tokens so I think this is really useful and necessary if we want the gem to be used widely.

petergoldstein:
@MarioRuiz If you need another option for handling multitenant OpenAI access you can check out the gem I created.  I don't want to abuse another repos comments, so I won't link directly, but it's easy enough to find.

alexrudall:
OK, let's do this - thanks for patience everyone. Working with it on a separate branch where I've done the rebase: #260

alexrudall:
Closing for #260

PR #153, opened 2023-01-06
Update the POST actions so that they use the correct Content-Type headers

petergoldstein:
This change also fixes the existing NPE on the cancel action.

I'm backporting these changes from my larger scale rewrite, because the latter is getting larger and diverting more substantially from the existing gem.

While this commit / PR doesn't contain any additional specs, I'd recommend that some unit tests be added (or pulled back from the rewrite branch) to ensure against regression.  The VCR-based specs are not sufficient as is.

This is an alternative to #142 , implementing the suggestion I made there.

It's worth noting that the only reason the existing specs are passing is because VCR is doing default matching - method and URI - and is not matching request headers or body.  So the content-type and the body are getting ignored.  I attempted to turn on matching there, but I got weird results (it looked like it was running new episodes) on my attempt so I let it go for the moment. 

## All Submissions:

### DIFF
--- a/lib/ruby/openai/client.rb
+++ b/lib/ruby/openai/client.rb
@@ -8,15 +8,15 @@ module OpenAI
     end
 
     def completions(parameters: {})
-      OpenAI::Client.post(path: "/completions", parameters: parameters)
+      OpenAI::Client.json_post(path: "/completions", parameters: parameters)
     end
 
     def edits(parameters: {})
-      OpenAI::Client.post(path: "/edits", parameters: parameters)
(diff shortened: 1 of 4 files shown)


alexrudall (approved):
@petergoldstein thanks a lot for taking the time to extract this, very helpful. I'll address the test issues separately.

alexrudall:
@petergoldstein you can now run the specs without VCR so they actually hit the API. This will cost about 2 cents and they should now all pass. You'll need to add an `OPENAI_ACCESS_TOKEN=` in `.env`.

```
  NO_VCR=true bundle exec rspec
```

PR #192, opened 2023-02-15
adds the ability to configure the timeout for the OpenAI requests

mbackermann:
This PR adds the ability to set the timeout (in seconds) for OpenAI requests. It can be configured by two ways:
1) Passing `request_timeout` named param when initializing the client. eg.
```ruby
OpenAI::Client.new(request_timeout: 25)
```

2) when configuring the gem. eg.
```ruby
OpenAI.configure do |config|
  config.request_timeout = 25
end
```

The default value is 60 seconds, which is the default value of the `Net::HTTP` class

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,24 @@ Then you can create a client like this:
     client = OpenAI::Client.new
 ```
 
+#### Setting request timeout
+
+The default timeout for any OpenAI request is 60 seconds. You can change that passing the `request_timeout` when initializing the client:
+
+```ruby
+    client = OpenAI::Client.new(access_token: "access_token_goes_here", request_timeout: 25)
+```
+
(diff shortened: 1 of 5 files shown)


codenamev (approved):
This looks great 💯 Much needed!

logicalgroove:
I am really waiting for this to be merged. For now, I'm grabbing the code from this pr.

ustincameron (approved):
Great PR, I'm also constantly hitting a timeout issue.

Can we get this merged in?

KonstantinKo:
Thank you very much! Works like a charm. I wonder if the default timeout shouldn't be higher, it seems like OpenAI frequently takes longer than 60 seconds to send more than a thousand tokens.

PR #217, opened 2023-03-15
Whisper example: files must be opened in binary mode

Tamnac:
Fixed the whisper example in readme.md. Files must be opened in binary mode, otherwise it will not work.

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ The translations API takes as input the audio file in any of the supported langu
 response = client.translate(
     parameters: {
         model: "whisper-1",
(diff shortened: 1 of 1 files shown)


alexrudall:
Hey @Tamnac! Thanks for this. Can you share an example of a file / format that would fail without this change? Currently we use `spec/fixtures/files/audio_sample.mp3`

alexrudall:
OK, makes sense - have updated the specs to match this also and seem to work fine.

alexrudall:
Thanks for your contribution!

PR #220, opened 2023-03-22
added configurable timeout to client

frmsaul:
I want to support configurable timeout option so I can set it to a pretty high number to avoid `Net::ReadTimeout with #<TCPSocket:(closed)> (Net::ReadTimeout)` errors


frmsaul:
Looks like this was implemented already

PR #230, opened 2023-03-27
Fix uri_base sample value in README

dlackty:
The current sample value in README actually doesn't work, and results in url like `oai.hconeai.comv1`. The trailing slash is essential.

## All Submissions:


alexrudall:
Thanks a lot @dlackty!

PR #214, opened 2023-03-10
test: allow running tests without secrets

bf4:
Since we already have VCR recording tests, it should be
possible to run the test suite without requiring the credentials,
since they're only required for live tests.

If I'm just picking up the gem for a quick PR, this lowers the barrier to entry, I think.

 ## All Submissions:

* [x] Have you followed the guidelines in our [Contributing document](../blob/main/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

PR #233, opened 2023-03-31
If token found, make specs always hit the actual API

alexrudall:
- Makes specs always hit the API if a token is found, either in ENV or passed in explicitly
- Allow specs to run even if the user does not have `OPENAI_ACCESS_TOKEN` in ENV, as suggested in [https://github.com/alexrudall/ruby-openai/pull/214](https://github.com/alexrudall/ruby-openai/pull/214)

Screenshot 2023-03-31 at 18 17 38

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -32,7 +32,7 @@ RSpec.configure do |c|
     warning = "WARNING! Specs are hitting the OpenAI API using your OPENAI_ACCESS_TOKEN! This
 costs at least 2 cents per run and is very slow! If you don't want this, unset
 OPENAI_ACCESS_TOKEN to just run against the stored VCR responses.".freeze
-    warning = RSpec::Core::Formatters::ConsoleCodes.wrap(warning, :red)
+    warning = RSpec::Core::Formatters::ConsoleCodes.wrap(warning, :bold_red)
 
     c.before(:suite) { RSpec.configuration.reporter.message(warning) }
     c.after(:suite) { RSpec.configuration.reporter.message(warning) }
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -26,9 +28,27 @@
     rspec.syntax = :expect
   end
 
+  if ENV.fetch("OPENAI_ACCESS_TOKEN", nil)
+    warning = "WARNING!
+    Specs are hitting the OpenAI API using your OPENAI_ACCESS_TOKEN!
+    This costs at least 2 cents per run and is very slow!
+    If you don't want this, unset OPENAI_ACCESS_TOKEN to just run against the stored VCR responses.
```
bf4, line 35: seems to me `ENV["NO_VCR"] == "true"` is the better way to toggle "please run live?"

spec/spec_helper.rb
```diff
@@ -26,9 +28,27 @@
...
+    ".freeze
+    colour_code = 31 # red
+    colour_warning = "\e[#{colour_code}m\n#{warning}\e[0m".freeze
```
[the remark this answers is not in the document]
alexrudall: Nice

spec/spec_helper.rb
alexrudall: Decided to try with OPENAI_ACCESS_TOKEN based on a couple things

- Simpler - just provide a token, and the tests will use it, even if a dev doesn't know what VCR is
- OpenAI change the API in a breaking way quite often, I want to run the specs vs the API more often in order to catch this quicker

Downsides

- Risk of costing money unexpectedly (mitigating with warnings)

If it causes problems I'll rethink.

alexrudall:
Thanks for the review @bf4 - much appreciated!

PR #234, opened 2023-04-02
Add streaming with Faraday

alexrudall:
- WIP on v4 of the gem - main breaking change is switching HTTP library to improve performance and allow streaming and parallel requests
- Switch to Faraday
- Add streaming to Chat as MVP


ruby-openai.gemspec
```diff
@@ -25,7 +25,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "httparty", ">= 0.18.1"
+  spec.add_dependency "faraday", "2.7.4"
+  spec.add_dependency "faraday-typhoeus", "1.0.0"
```
bf4, line 29: I personally have no interest in using or install typhoeus on my system.  The advantage of Faraday is you can leave those decisions to someone else.

If you want to test Faraday with typhoeus in the gem, the pattern for that would be to put typhoeus in the Gemfile but not the gemspec, that way it's not required.

Also, our Faraday requirement in our app

```
gem "faraday", ">= 2"
```

We happen to be on 2.7.4, but I'd prefer not to be locked in to that version

```diff
-  spec.add_dependency "faraday", "2.7.4"
+  spec.add_dependency "faraday", ">= 2"
```

lib/openai/client.rb
```diff
@@ -52,37 +52,39 @@ def translate(parameters: {})
...
     def self.multipart_post(path:, parameters: nil)
-      HTTParty.post(
-        uri(path: path),
-        headers: headers.merge({ "Content-Type" => "multipart/form-data" }),
-        body: parameters,
-        timeout: request_timeout
-      )
+      conn.post(uri(path: path), timeout: OpenAI.configuration.request_timeout) do |req|
+        req.body = parameters
+        req.headers = headers.merge({ "Content-Type" => "multipart/form-data" })
```
bf4, line 74: you might want faraday-multipart to take advantage of multipart responses

lib/openai.rb
```diff
@@ -17,7 +18,7 @@ class Configuration
 
     DEFAULT_API_VERSION = "v1".freeze
     DEFAULT_URI_BASE = "https://api.openai.com/".freeze
-    DEFAULT_REQUEST_TIMEOUT = 120
+    DEFAULT_REQUEST_TIMEOUT = 0 # Never time out.
```
bf4, line 21: given how unstable the OpenAI API is, this doesn't seem a great default... I personally would prefer a much shorter than 120 seconds request timeout.

nicastelo: Agree with @bf4. 120 seconds seems fine to me though

codenamev:
This looks great!  How can I help push this through the finish line?

ruby-openai.gemspec
alexrudall: Great points, thanks for review! Experimenting with Faraday only version.

alexrudall:
Streaming getting there :) 

https://user-images.githubusercontent.com/7175262/233973159-2493cf81-bb9d-4048-bbc3-3fe2a3ea4ac5.mp4

ruby-openai.gemspec
```diff
@@ -25,7 +25,9 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "httparty", ">= 0.18.1"
+  spec.add_dependency "faraday", ">= 2"
```
fwitzke, line 28: @alexrudall iIs there an actual requirement on faraday `2`, or could this dependency be relaxed to allow older versions of faraday? I'm asking because I wanted to test this branch out in an existing codebase that uses an older version of faraday, and since it wasn't easy to upgrade everything else to faraday 2 I just tested with `1.10` and everything seemed worked fine (at least for my use case).
```suggestion
  spec.add_dependency "faraday", "< 3"
```

alexrudall: No requirement that I can see, changed to >= 1

ruby-openai.gemspec
```diff
@@ -25,7 +25,9 @@ Gem::Specification.new do |spec|
...
+  spec.add_dependency "faraday-multipart", ">= 1"
+  spec.add_dependency "mime-types", ">= 3"
```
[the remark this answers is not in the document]
alexrudall: Good point, removed

alexrudall:
Thanks for your input all. Would appreciate any final reviews 👍 

- This PR switches the HTTP library from HTTParty to Faraday and adds streaming functionality on the Chat endpoint. 
- Breaking change: endpoints now return JSON rather than HTTParty objects. You need to update your code to handle this change, changing code like `JSON.parse(response.body)["key"]` and `response.parsed_response["key"]` to just `response["key"]`.
- README examples should all be updated and working and including Chat streaming.

lib/openai/client.rb
```diff
@@ -52,37 +52,62 @@ def translate(parameters: {})
...
+      JSON.parse(string.gsub("}\n{", "},{").prepend("[").concat("]"))
+    end
+
+    private_class_method def self.to_json_stream(user_proc:)
+      proc do |chunk, bytesize|
+        # Clean up response string of chunks.
+        chunk = chunk.gsub("\n\ndata: [DONE]\n\n", "").gsub("\n\ndata:", ",").gsub("data: ", "")
+
+        # Turn it into JSON.
+        chunk = to_json(chunk)
```
fwitzke, line 102: @alexrudall I tried this branch locally and it's working for the most part, but I'm occasionally seeing the following error thrown in this `to_json` call:
```ruby
451: unexpected token at 'DONE]\n\n]'
```

It seems like the `gsub` call above might be missing some cases, although I wasn't able to get to the root cause yet, I'll take another look and see if I can consistently reproduce this but just wanted to raise in case you've seen this error (also see you pushed some new commits so it might have been fixed already).

rmontgomery429: Instead of the calls to `gsub` it seems like it would only be necessary to capture data if it looks like a JSON object. You could do someting like `json_data = chunk.match(/data: (\{.*\})/i)` and only attempt to parse the data if `json_data` is present. If it's not, it would be `[DONE]` or some other non-json like object. I think their API docs suggest it should either be a JSON object, i.e. `{...}` or `[DONE]` so those chunks that do not match could be ignored. This would simplify this code a bit and handle the situation found by @fwitzke.

rmontgomery429: I'm working on a writing some specs and was able to reproduce 

```
JSON::ParserError:
       unexpected token at 'foo]'
```

Basically just pass "foo" as a chunk and it will try to parse it, fail, and then attempt to turn it into an array of JSON objects which is where the `]` comes from. I'll try to wrap this up today and submit for review.

fwitzke: Also while trying to repro this I got one error response from openai API, it doesn't seem like the code handles that right now, this is the `chunk` data received:
```
data: {"id":"chatcmpl-79HdSUgAA1wTHnw8fKIboQcadubqK","object":"chat.completion.chunk","created":1682447098,"model":"gpt-3.5-turbo-0301","choices":[{"delta":{},"index":0,"finish_reason":"stop"}]}

data: {"error":{"message":"The server experienced an error while processing your request. Sorry about that! You can retry your request, or contact us through our help center at help.openai.com if the error persists.","type":"server_error","param":null,"code":null}}

data: [DONE]

```

rmontgomery429: For what it's worth I did make an attempt here: https://github.com/alexrudall/ruby-openai/pull/248. Your feedback is very welcome and I'm happy to adjust if I'm on an agreeable track.

fwitzke: that simplifies a lot @rmontgomery429, I think that change makes sense and would address both cases I mentioned!

lib/openai/client.rb
```diff
@@ -52,37 +52,62 @@ def translate(parameters: {})
...
+
+        # Pass an array of JSONified chunk(s) to the user's Proc.
+        user_proc.call([chunk].flatten, bytesize)
+      end
+    end
+
+    private_class_method def self.conn
+      Faraday.new
```
fwitzke, line 110: I'm not sure if it's specific to the `faraday` version I'm using, but it looks like the `post` calls (probably all calls) don't actually accept `timeout` as an option, so that is not being applied. I was able to keep the timeout setting by doing this instead of passing `timeout` as an argument to post/get etc:
```suggestion
      Faraday.new do |f|
        f.options[:timeout] = OpenAI.configuration.request_timeout
      end
```

lib/openai/client.rb
alexrudall: Thanks for the excellent work on this. Merged without comment!

lib/openai/client.rb
alexrudall: Nice spot, have fixed this and changed the tests so they'll catch this issue in future.

kmcphillips:
This is rad. Thanks for this change! :heart:

gastonmorixe:
Thank you everyone!

lib/openai/client.rb
```diff
@@ -52,33 +52,42 @@ def translate(parameters: {})
...
+      end)
+    end
+
+    def self.json_parse(response)
+      return unless response
+
+      JSON.parse(response.body)
+    rescue JSON::ParserError
+      # Convert a multiline file of JSON objects to a JSON array.
+      JSON.parse(response.body.gsub("}\n{", "},{").prepend("[").concat("]"))
```
andresgutgon: Hi @alexrudall, I come here now in October 😅. I was wondering when happens this case. I have seen that this happen to us when OpenAI returns empty string. Is this documented somewhere why it happens?

PR #238, opened 2023-04-11
Add Microsoft Azure support

### LINKED ISSUE #231: Microsoft Azure support
**Is your feature request related to a problem? Please describe.**

We would like to use `ruby-openai` with Microsoft Azure OpenAI, however its endpoints and authentication are handled slightly differently, making it impossible currently to use this gem.

**Describe the solution you'd like**

I propose to support a configuration setting `api_type`, which can then be set to `:azure`, similar to how https://github.com/openai/openai-python supports Microsoft Azure.

**Describe alternatives you've considered**

I'm aware of the changes being worked on in #150, it might be possible to create a subclass of `OpenAI::Client` once this has been implemented.

**Additional context**

Happy to submit a PR if desired!

rmachielse:
**Problem**

Resolves #231. We would like to use `ruby-openai` with Microsoft Azure OpenAI, however its endpoints and authentication headers are slightly different, making it impossible currently to use this gem.

**Solution**

This PR adds a configuration setting `api_type`, which can be set to `:azure`, similar to how https://github.com/openai/openai-python supports Microsoft Azure.

**Alternatives**

I'm aware of the changes being worked on in https://github.com/alexrudall/ruby-openai/pull/150, it might be possible to create a subclass of `OpenAI::Client` instead once this has been implemented. I'm happy to rebase and refactor this PR once this is merged.

**Additional context**

- Microsoft Azure currently only supports completions and embeddings, so these are the only endpoints implemented
- A Microsoft Azure account is needed to rerecord the tests with VCR, and the account needs to have OpenAI access. Currently a form needs to be submitted for this. Also, a deployment has to be created for every model used in the tests, and the deployment name has to match the model name. Documentation should probably be written for this.

### DIFF
--- a/lib/openai.rb
+++ b/lib/openai.rb
@@ -15,7 +15,7 @@ module OpenAI
 
   class Configuration
     attr_writer :access_token
-    attr_accessor :api_version, :organization_id, :uri_base, :request_timeout
+    attr_accessor :api_type, :api_version, :organization_id, :uri_base, :request_timeout
 
     DEFAULT_API_VERSION = "v1".freeze
     DEFAULT_URI_BASE = "https://api.openai.com/".freeze
@@ -23,6 +23,7 @@ module OpenAI
 
     def initialize
       @access_token = nil
+      @api_type = nil
       @api_version = DEFAULT_API_VERSION
       @organization_id = nil
       @uri_base = DEFAULT_URI_BASE
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -68,14 +68,27 @@ module OpenAI
     end
 
     def uri(path:)
-      OpenAI.configuration.uri_base + OpenAI.configuration.api_version + path
+      if OpenAI.configuration.api_type == :azure
+        OpenAI.configuration.uri_base + path + "?api-version=#{OpenAI.configuration.api_version}"
+      else
+        OpenAI.configuration.uri_base + OpenAI.configuration.api_version + path
+      end
     end
 
     def headers
+      return azure_headers if OpenAI.configuration.api_type == :azure
+
       {
         "Content-Type" => "application/json",
         "Authorization" => "Bearer #{OpenAI.configuration.access_token}",
         "OpenAI-Organization" => OpenAI.configuration.organization_id
+      }
+    end
+
+    def azure_headers
+      {
+        "Content-Type" => "application/json",
+        "api-key" => OpenAI.configuration.access_token
       }
     end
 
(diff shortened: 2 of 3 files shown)


rmachielse:
Now sure why circle is failing, it seems all the tests are missing the `OPENAI_ACCESS_TOKEN`

steffansluis:
In cooperation with my colleague @rmachielse I've followed up on this. Initially I was looking at rebasing his changes, but I realized it would likely be easier to add the minimal required changes on top of `main` and slightly change the associated configuration changes.

In practice, the only differences that are really needed are to the `uri` and the `headers`, so I only ported those changes and added some tests for them. This works with the following (or similar) configuration (the example specifies `config/initializers/openai.rb`):
```ruby
require 'openai'

OpenAI.configure do |config|
  config.access_token = Rails.application.config.app[:azure_openai_api_key]
  config.uri_base = "#{Rails.application.config.app[:azure_openai_uri]}/openai/deployments/gpt-35-turbo"
  config.api_type = :azure
  config.api_version = '2023-03-15-preview'
end
```
where `azure_openai_uri` is something like `https://custom-domain.openai.azure.com`.

In the process I realized there was an inconsistency with `/`es in the constructed URI, which I tried to fix with `URI.join`, however this resulted in some code constructions I didn't like so I ended up removing that part of the logic again. That unfortunately leads to the inconsistency where when using Azure, no trailing `/` is expected in the `uri_base` but when using the regular API a trailing `/` is expected and necessary.

I think the ability to be able to use Azure outweighs this inconsistency so I went ahead with the changes like this. If you want I can see if I can come up with a clean way of addressing this. My suggestion would be to do it in a separate PR though, since the behavior was already problematic (no trailing `/` results in problems with the current logic) and being able to use the gem with Azure outweighs delaying fixing this in my opinion.

Curious to hear what you think @alexrudall !

pickhardt:
Looks good to me. By comparison, this is what the api would look like in Python:

```
import openai
openai.api_type = "azure"
openai.api_base = "https://your-resource-here.openai.azure.com/"
openai.api_version = "2023-03-15-preview"
openai.api_key = "your-key-here"
```

rmachielse:
@alexrudall what do you think? These minimal changes should make it quite straight forward to at least allow usage of this gem with Microsoft Azure.

timrwilliams:
I'd also like to see this merged. I think more and more people will be transitioning across to the Azure version, especially within corporate environments so this would be very helpful.

IanRAnderson:
I'm also hoping for this functionality. Been using this Gem for developing a proof of concept for a new feature, but my organization is moving to Azure OpenAI for a variety of reasons, so the ability to continue to use this gem with Azure would be great.

eichert12:
We're also moving to an Azure based version of OpenAI so would love to see Azure support in this gem.

PR #248, opened 2023-04-25
Improved stream parsing

rmontgomery429:
# Overview of changes

This PR will change how streaming parses chunks of data. Instead of "cleaning" up the data and parsing it as a large multi-line chunk, it reads data chunks by scanning with a regular expression and for each capture it passes that to the users proc.

## All Submissions:

### DIFF
--- a/spec/openai/client/client_spec.rb
+++ b/spec/openai/client/client_spec.rb
@@ -95,7 +95,9 @@ RSpec.describe OpenAI::Client do
 
         it "does not raise an error" do
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
-          expect(user_proc).to_not receive(:call).with(JSON.parse('{"message": "A bad thing has happened!"}'))
+          expect(user_proc).to_not receive(:call).with(
+            JSON.parse('{"message": "A bad thing has happened!"}')
+          )
 
           expect do
             stream.call(chunk)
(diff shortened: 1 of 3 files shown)


spec/openai/client/client_spec.rb
```diff
@@ -25,6 +25,88 @@
...
+            stream.call(chunk)
+          end.to_not raise_error(JSON::ParserError)
+        end
+      end
+
+      context "when called with a string containing an error" do
+        let(:chunk) do
+          <<-CHUNK
+            data: { "foo": "bar" }
+            error: { message: "A bad thing has happened!" }
```
fwitzke, line 92: I think this is fine in the spec context to make sure it filters out anything that doesn't look like valid data, but I think errors would also be sent as `data: {}` though, see https://github.com/alexrudall/ruby-openai/pull/234#discussion_r1176889900

which is not problem necessarily as it allows callers to handle those errors in the same callback

rmontgomery429: Yes. This is an open question. There are two kinds of errors here that seem like they will come up often. Parsing errors and just bad data seems less likely but possible. Then there are the "good" errors coming from OpenAI where they return an `error: ...` message. What should happen to those "good" error messages? It seems like it would be a good idea to yield those errors to the user_proc for now as JSON so the user can do whatever they would like with them (skip, try again, etc.). I'm not sure if skipped, the next chunk of valid data would pick back up where it left off or not.

rmontgomery429: @fwitzke I decided to send the OpenAI errors to the user proc.

alexrudall (approved):
This is really great stuff, thanks so much!

Issue #251, opened 2023-04-28
Incomplete response with streaming 


deikka:
I have a chat system working perfectly. When I updated to version 4 of the gem, everything went well (it works normally). However, when I add the 'stream' option to the OpenAI API call, the content of the response is incomplete.

This code works normally:
```ruby
response = openai_client.chat(
  parameters: {
    model: "gpt-4",
    messages: [
      {role: "system", content: system_prompt},
      {role: "user", content: user_prompt}
    ],
    temperature: 0.4,
    user: "user_#{user_id}"
  }
)
puts "RESPONSE: #{response}"
response.dig("choices", 0, "message", "content")
``` 
However, when adding the streaming option:
```ruby
response = openai_client.chat(
  parameters: {
    model: "gpt-4",
    messages: [
      {role: "system", content: system_prompt},
      {role: "user", content: user_prompt}
    ],
    temperature: 0.4,
    stream: proc do |chunk, _bytesize|
      new_content = chunk.dig("choices", 0, "delta", "content")
      if new_content
        answer.content = (answer.content || "") + new_content
        answer.save!
      end
    end,
    user: "user_#{user_id}"
  }
)
puts "RESPONSE: #{response}"
response.dig("choices", 0, "message", "content")
``` 
Screenshot with responses:
![Captura de pantalla 2023-04-28 a las 12 29 24](https://user-images.githubusercontent.com/13509363/235124676-af040299-8456-4c73-be7f-7cc88935a24b.png)

Any clue as to why it could happen?

- Rails 7.0.4.3
- Ruby 3.2.1
- OS: macOS
- Browser ARC/Chrome

alexrudall:
Thanks for sharing on here! I don't have gpt-4 access and I don't know what your system prompt is, with gpt-3.5 I'm not able to reproduce this so far:

```
openai_client.chat(
  parameters: {
    model: "gpt-3.5-turbo",
    messages: [
      {role: "system", content: "Answer politely."},
      {role: "user", content: "What is the maximum weight allowed in the closed drawer or in combination of inside and on top of the open drawer?"}
    ],
    temperature: 0.4,
    stream: proc do |chunk, _bytesize|
      print chunk.dig("choices", 0, "delta", "content")
    end,
  }
)
#=> I'm sorry, but I don't have enough information to answer your question. Could you please specify which drawer or product you are referring to?
```

Are you able to print out all chunks and see if they are coming through correctly?

rmontgomery429:
[...]

PR #252, opened 2023-04-28
Add a spec to test the client with GPT-4

rmontgomery429:
## All Submissions:

This adds a spec to test the client using gpt-4 as it was mentioned that @alexrudall does not have access to gpt-4 yet in #251. The problem was not manifested but it's still nice to have this spec for sake of completeness.


alexrudall:
Thanks for your work on this! Couple of things with this

- I like the shared examples and it's nice to test the separate models, as sometimes OpenAI remove them without much warning and good to catch that so we can update the README 
- Would be great to reproduce the Helicone issue but that's a separate thing maybe and also might be hard as the responses from the AI is not idempotent so we need to somehow check it's speaking correct English without missing any words? lol. I thought I read somewhere the chunks returned an index but all these have index: 0. Maybe they're coming through as one chunk and if they came separately like they seem to in production we'd be able to see the issue?
- I need anyone who has an API key to be able to run the full specs (with the non-VCR option) and them pass, because it feels more inclusive and will also annoy me if that spec always fails for me until I get through the waitlist. unfortunately that means leaving GPT-4 out for now.

Anyway if you remove the GPT-4 spec (for now) I'll merge. BTW, you're the first contributor I've seen who's understood how I've written the tests and how to add to them (and actually improve them) and I really appreciate that.

rmontgomery429:
@alexrudall I'm happy to help. Regarding GPT-4, I agree with being more inclusive, but I also think like it's important to be able to prove that the gem works with as much of the OpenAI API as possible. Would you consider a switch to opt-in to testing with GPT-4? It could be something simple like setting an environment variable `GPT4=true`?

rmontgomery429:
@alexrudall Check out the latest changes and let me know what you think. I've added a note to the readme as well.

README.md
```diff
@@ -364,6 +364,10 @@ OPENAI_ACCESS_TOKEN=123abc bundle exec rspec
 
 Then update the version number in `version.rb`, update `CHANGELOG.md`, run `bundle install` to update Gemfile.lock, and then run `bundle exec rake release`, which will create a git tag for the version, push git commits and tags, and push the `.gem` file to [rubygems.org](https://rubygems.org).
 
+## Testing GPT-4
+
+Since GPT-4 isn't widely available yet, and this project aims to be as includsive as possible, specs that test GPT-4 models require an additinoal enviroment variable to be set. To test GPT-4 models simply set `GPT4=true` in your environment or on the command line like this: `GPT4=true bundle exec rspec`.
```
kykocamp, line 369: **suggestion (non-blocking):** 

```suggestion
Since GPT-4 isn't widely available yet, and this project aims to be as inclusive as possible, specs that test GPT-4 models require an additinoal environment variable to be set. To test GPT-4 models simply set `GPT4=true` in your environment or on the command line like this: `GPT4=true bundle exec rspec`.
```

jamiemccarthy, line 369:
```suggestion
Since GPT-4 isn't widely available as of July 2023, and this project aims to be as inclusive as possible, specs that test GPT-4 models will not run unless an environment variable is set. To test GPT-4 models, set `GPT4=true` in your environment or on the command line like this: `GPT4=true bundle exec rspec`.
```

rmontgomery429:
This has gone stale.

PR #260, opened 2023-05-12
Support multi-tenancy

alexrudall:
- Continuing from #150 on a new branch 
- Removed the NULL_ORGANIZATION_ID - I think having a default of falling back to global config organization_id is OK for now.
- Removed api_version as there is only 1 api version currently, can add it when needed.
- As suggested by Peter in his PR, removed ability to pass access_token to other classes like Files . You must pass a client.
- This is a breaking change (as it will no longer be possible to e.g. call Files#list with an access token) and will be a major version bump.

### With Config

For a more robust setup, you can configure the gem with your API keys, for example in an `openai.rb` initializer file. Never hardcode secrets into your codebase - instead use something like [dotenv](https://github.com/motdotla/dotenv) to pass the keys safely into your environments.

```ruby
OpenAI.configure do |config|
    config.access_token = ENV.fetch("OPENAI_ACCESS_TOKEN")
    config.organization_id = ENV.fetch("OPENAI_ORGANIZATION_ID") # Optional.
end
```

Then you can create a client like this:

```ruby
client = OpenAI::Client.new
```

You can still override the config defaults when making new clients; any options not included will fall back to any global config set with OpenAI.configure. e.g. in this example the organization_id, request_timeout, etc. will fallback to any set globally using OpenAI.configure, with only the access_token overridden:

```ruby
client = OpenAI::Client.new(access_token: "access_token_goes_here")
```


pickhardt:
I know multi-tenancy isn't directly the same as Microsoft Azure support, but I saw the other pull request was discussing this PR so I am leaving a comment here. Is this PR also supposed to let you use Azure by configuration the endpoint and/or other configuration values?

If so, I think you should give an example in the readme for how you can use these changes with Microsoft Azure, which will probably be a pretty common use case. I was comparing this PR with https://github.com/alexrudall/ruby-openai/pull/238/files because that other PR changes the headers a bit. I think this PR doesn't support the Azure use case since you can't change the headers with configuration values?

cthulhu:
Need any help testing this bit?

alexrudall:
> Need any help testing this bit?

Yes - would you mind giving it a spin? If you OK I'll merge :) Thanks so much @cthulhu (great handle)

cthulhu:
Ok installed it from git and branch:
   
    Using ruby-openai 4.1.0 (was 4.2.0) from https://github.com/alexrudall/ruby-openai.git (at support_multi_tenancy@ecafc47)
    
OpenAI default API works. 

There is a small drift between this branch and the main, regarding api_type. So I could not create an Azure client there. My use case is that I'm using both providers in the same app. I use 3 clients in the same Ruby app:

- OpenAI client
- Azure Chat (France region)
- Azure Embeddings (West EU region)

Will be happy to help with merge/conflicts and further tests.

alexrudall:
@cthulhu thanks so much for testing this. I've now rebased and it should be good to go. If you have time I would really appreciate one more test before I merge.

Issue #256, opened 2023-05-05
No exception information returned in Stream mode


lingceng:
The api returns a JSON when error, eg. 

```
{
  "error": {
    "message": "'user_role' is not one of ['system', 'assistant', 'user'] - 'messages.0.role'",
    "type": "invalid_request_error",
    "param": null,
    "code": null
  }
}
```

I use following  code to fix

```
    def self.to_json_stream(user_proc:)

      proc do |chunk, overall_received_bytes, env|
        Rails.logger.debug("raw chunk: #{chunk}")

        if chunk.start_with?("data: {")
          chunk.each_line do |line|
            next unless line.start_with?("data: {")
            line.delete_prefix!("data: ")
            user_proc.call(JSON.parse(line))
          end
        elsif chunk.include?('"error": ')
          user_proc.call(JSON.parse(chunk))
        else
          Rails.logger.warn("unkown raw chunk: #{chunk}")
        end
      end
    end
```

lingceng:
The stream data may be split, here is the complete solution: https://gist.github.com/lingceng/afd30be6e102442fa48abf175efe3f6a

Any one can help merge to the project.

@itsderek23 @obie

harlantwood:
FYI I have an open PR to solve the same issue, using a different approach: https://github.com/alexrudall/ruby-openai/pull/275

atesgoral:
I'm working on properly fixing these issues with #338 (and I'll follow up with a breaking change to raise actual exceptions)

alexrudall:
Released in [v6](https://github.com/alexrudall/ruby-openai/releases/tag/v6.0.0)

PR #270, opened 2023-05-28
Don't swallow model errors in streaming mode

obie:
## All Submissions:

I lost a lot of time and suffered needless premature aging until putting this change into place on my project. A bug was causing context overflows and the current release of the library swallows those errors.

Credit to @lingceng for the solution in https://github.com/alexrudall/ruby-openai/issues/256


lib/openai/http.rb
```diff
@@ -53,11 +53,19 @@ def to_json(string)
...
+      proc do |chunk, overall_received_bytes, env|
+        Rails.logger.debug("raw chunk: #{chunk}")
+
+        if chunk.start_with?("data: {")
+          chunk.each_line do |line|
+            next unless line.start_with?("data: {")
+            line.delete_prefix!("data: ")
+            user_proc.call(JSON.parse(line))
+          end
+        elsif chunk.include?('"error": ')
```
igorkasyanchuk, line 65: don't know 100% how it works, but can it just return me some JSON examples with this string?

Pklong:
```
raw chunk: {
    "error": {
        "message": "The model `gpt-3.5-turbo!!!!` does not exist",
        "type": "invalid_request_error",
        "param": null,
        "code": null
    }
}
```

harlantwood:
FYI I have an open PR to solve the same issue, using a different approach: https://github.com/alexrudall/ruby-openai/pull/275

lib/openai/http.rb
```diff
@@ -53,11 +53,19 @@ def to_json(string)
     # @param user_proc [Proc] The inner proc to call for each JSON object in the chunk.
     # @return [Proc] An outer proc that iterates over a raw stream, converting it to JSON.
     def to_json_stream(user_proc:)
-      proc do |chunk, _|
-        chunk.scan(/(?:data|error): (\{.*\})/i).flatten.each do |data|
-          user_proc.call(JSON.parse(data))
-        rescue JSON::ParserError
-          # Ignore invalid JSON.
+      proc do |chunk, overall_received_bytes, env|
+        Rails.logger.debug("raw chunk: #{chunk}")
```
codenamev, line 57: Please remove references to `Rails`.  It's not a dependency and will cause issues for anyone using this outside of Rails.

I've created a new issue (#320) to track adding a logger, and this can be added there.

atesgoral:
I'm working on properly fixing these issues with #338 (and I'll follow up with a breaking change to raise actual exceptions)

alexrudall:
Thanks @obie -- going with #344 for this

PR #287, opened 2023-06-27
Skip flatten.each in HTTP to_json_stream

Maumagnaguagno:
Process each scanned element from chunk without intermediate array created by scan, flattened and iterated again.

## All Submissions:

### DIFF
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -54,7 +54,7 @@ module OpenAI
     # @return [Proc] An outer proc that iterates over a raw stream, converting it to JSON.
     def to_json_stream(user_proc:)
       proc do |chunk, _|
-        chunk.scan(/(?:data|error): (\{.*\})/i).flatten.each do |data|
+        chunk.scan(/(?:data|error): (\{.*\})/i) do |data,|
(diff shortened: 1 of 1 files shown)


alexrudall:
Thanks @Maumagnaguagno - this seems to work but I can't find the Ruby docs for the mechanism here, can you explain this a bit / link to them? Thank you!

Maumagnaguagno:
[String#scan](https://ruby-doc.org/core-3.1.0/String.html#method-i-scan) without a block returns an arrays of match arrays, even if there is a single capturing group/matched element.
The [Array#flatten](https://ruby-doc.org/core-3.1.0/Array.html#method-i-flatten) is currently used to remove the array around the matched element.
By passing block to String#scan with ``|data,|`` we can splat the matched array just like flatten, keeping only data as the first element, without the intermediate arrays created by scan and flatten.
Another solution is to avoid the capturing group and prefer a lookbehind regex.

alexrudall:
Using the new [event_stream_parser](https://github.com/Shopify/event_stream_parser) for this. Thanks anyway for this contribution!

PR #300, opened 2023-08-11
Support multiple deployment_ids for azure

henrikbjorn:
This enables support for calling multiple azure deployments when using the library. This follows the same convention as the official python openai package.

So `client.embeddings(deployment_id: 'my-gpt-deployment', ...rest_of_args)`

Since a embedding model would be one deployment and the actual chat model would be another this is required to use both.

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,19 @@ To use the [Azure OpenAI Service](https://learn.microsoft.com/en-us/azure/cognit
     end
 ```
 
-where `AZURE_OPENAI_URI` is e.g. `https://custom-domain.openai.azure.com/openai/deployments/gpt-35-turbo`
+where `AZURE_OPENAI_URI` is e.g. `https://custom-domain.openai.azure.com`
+
+Then add the `deployment_id` to the api calls. e.g:
+
+``` ruby
+client.embeddings(
+  deployment_id: 'gpt-35-turbo',
+  parameters: {
+    model: 'text-embedding-ada-002',
+    input:
+  }
+)
+```
 
 ### Models
 
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -6,7 +6,9 @@ module OpenAI
       end&.body)
     end
 
-    def json_post(path:, parameters:)
+    def json_post(path:, parameters:, deployment_id: nil)
+      path = "/openai/deployments/#{deployment_id}/#{path}" if deployment_id
+
       to_json(conn.post(uri(path: path)) do |req|
         if parameters[:stream].respond_to?(:call)
           req.options.on_data = to_json_stream(user_proc: parameters[:stream])


henrikbjorn:
> Thanks for this. Looks like a breaking change for Azure users?

Not really, they can just use the same base uri as always and then omit the `deployment_id` parameter from their method calls.

henrikbjorn:
Sorry to ping you @alexrudall 

Not sure what you want to do with the Rubycop warning

alexrudall:
@henrikbjorn - this should now be solved in [v5](https://rubygems.org/gems/ruby-openai/versions/5.0.0) of ruby-openai as you can create multiple Client objects each with different configuration (so you can set different URI as needed). Let me know if this doesn't solve your problem, and thank you for your contribution!

henrikbjorn:
> @henrikbjorn - this should now be solved in [v5](https://rubygems.org/gems/ruby-openai/versions/5.0.0) of ruby-openai as you can create multiple Client objects each with different configuration (so you can set different URI as needed). Let me know if this doesn't solve your problem, and thank you for your contribution!

That dosen't solve the problem. It makes it complex and unneeded. Since I would have to have a Embedding Client, GPT-4 Client etc.

That does nok make sense to me

henrikbjorn:
Since especially that usecase is even still supported with this code. Just set the base uri to the deployment directly and omit the deployment_id from the api calls

alexrudall:
Hmm OK let me take another look

henrikbjorn:
The `path` in `json_post` in only changed if `deployment_id` is present. Otherwise the code path is as before

themire:
This is what I'm doing to support multiple models in Azure which is working fine.

```ruby
module OpenAi
  class Client
    def self.client(deployment_id)
      OpenAI::Client.new(
        uri_base: "https://***.openai.azure.com/openai/deployments/#{deployment_id}"
      )
    end

    def self.default
      client("gpt-35-turbo")
    end

    def self.gpt35_16k
      client("gpt-35-16k")
    end

    def self.gpt4
      client("gpt-4-8k")
    end

    def self.embedding
      client("text-embedding-ada-002")
    end
  end
end



OpenAi::Client.gpt35_16k.chat(parameters: ...)
```

henrikbjorn:
Either works, but I think it is a good idea to follow the same API as the python SDK, so the SDKs works the same

bmulholland:
One minor annoyance with the current setup is that, even if creating multiple OpenAI clients, we have to duplicate and construct the uri each time. In that sense, the configuration for Azure isn't actually "URI base," but "full URI." Ideally, we'd configure the true "base" of the URI once (e.g. from an env var), and then set the deployment per client initialization. As a compromise, could we instead split out an optional deployment_id constructor arg that gets appended to the uri_base?

henrikbjorn:
> One minor annoyance with the current setup is that, even if creating multiple OpenAI clients, we have to duplicate and construct the uri each time. In that sense, the configuration for Azure isn't actually "URI base," but "full URI." Ideally, we'd configure the true "base" of the URI once (e.g. from an env var), and then set the deployment per client initialization. As a compromise, could we instead split out an optional deployment_id constructor arg that gets appended to the uri_base?

Which is what this PR does

PR #304, opened 2023-08-12
Fixes #259

### LINKED ISSUE #259: Transcribe throwing exception, expecting UploadFile
**Describe the bug**
When using V4.0.0 of the gem and attempting to pass a file to the `transcribe` it receives the response below. Downgrading to version 3.7.0 works as a workaround.

```ruby
{"error"=>{"message"=>"1 validation error for Request\nbody -> file\n  Expected UploadFile, received: <class 'str'> (type=value_error)", "type"=>"invalid_request_error", "param"=>nil, "code"=>nil}}
```

**To Reproduce**
1. Attempt to pass a file to the `transcribe` methods `file` parameter
2. See error above as the `response`

**Expected behavior**
It should allow the file to be passed to the service and not attempt to stringify it.

**Code**
```ruby
recording.audio.attachment.blob.open do |file|
  response = client.transcribe(parameters: { model: "whisper-1", file: file })
  recording.update(transcription: response.parsed_response['text'])
end
```

**Additional context**
It seems to be coming from https://github.com/alexrudall/ruby-openai/blob/main/lib/openai/http.rb#L84 as `StringIO`, `Tempfile` and others `is_a?(File)` is `false`.

codergeek121:
Thank you for the amazing gem 👍

This PR fixes the transcribe endpoint for Tempfiles.

Before this change, Tempfiles did not get transformed to Faraday::UploadIOs, since they do not pass the is_a?(File) check.

I actually tested this on a local Rails project with ActiveStorage (which mostly uses Tempfiles) and it works as expected. Unfortunately I did not manage to make the newly added spec pass, because of some VCR issues. I'm not deeply familiar with VCR, but I'll try to have another look tomorrow. Running the suite did change the transcribe cassette, which I did not expect, but maybe I'm misunderstanding VCR


alexrudall:
> Thank you for the amazing gem 👍
> 
> 
> 
> This PR fixes the transcribe endpoint for Tempfiles.
> 
> 
> 
> Before this change, Tempfiles did not get transformed to Faraday::UploadIOs, since they do not pass the is_a?(File) check.
> 
> 
> 
> I actually tested this on a local Rails project with ActiveStorage (which mostly uses Tempfiles) and it works as expected. Unfortunately I did not manage to make the newly added spec pass, because of some VCR issues. I'm not deeply familiar with VCR, but I'll try to have another look tomorrow. Running the suite did change the transcribe cassette, which I did not expect, but maybe I'm misunderstanding VCR

Well thank you for your contribution! This looks good, can you explain the context for passing a Tempfile? Is it streaming audio as you record it or something?

codergeek121:
I'll go into details what I was trying to do (and I'm currently writing a blogpost about it), feel free to ask if you're missing something:

I have a form in a Rails with ActiveStorage where users can upload their audiofiles. After the upload, ActiveStorage analyzes the audiofiles and adds metadata to the file's blob record, like length, bitrate etc. To get this metadata, ActiveStorage has to download the file from an external service (e.g. S3) and e.g. run `ffmpeg` against it. Instead of using `File` for this, it uses `Tempfile` to have the file automatically cleaned up after the analysis has been done. I augmented this analysis with a transcription API call, to save the text of audiofiles along with it.
ActiveStorage uses Tempfiles for the downloaded files, since it wants to make sure the files get cleaned up properly after doing operations on them and it probably does not care about filenames for these temporary files. My understanding of Tempfiles is, that they are like 99% identical to `File`s, but they have defaults for filenames and some cleanup logic.

```ruby
class Tempfile < DelegateClass(File)
```
This is the way Tempfiles are implemented, which means the `file.is_a? File` check, will not find the File class in the ancestors, even though it delegates nearly everything to it.

Another (maybe better) approach could be to change the conditional and remove both `is_a?` checks and replace them with `value.responds_to? :path` or maybe `:read`, since in the end we do not care about concrete class the file is implemented but if it works with the `UploadIO` class :)

alexrudall:
Thanks for that. Made some changes - specs pass locally but not on CI for some reason. Will try and fix.

alexrudall:
Tempfiles can be a slightly different length in different OSs it seems lol, so my hacky length + URI matching for VCR doesn't work on the linux distro on CI since I recorded the cassette on macOS. We do not need a test for this this bad, it works, deleted the test, deploying, 🙏

codergeek121:
Seems reasonable! Thank you for your help 🥇

alexrudall:
No problemo, released in [v4.3.1](https://rubygems.org/gems/ruby-openai/versions/4.3.1)

alexrudall:
Send that blogpost my way when it's done and I'll post on socials etc.

Issue #291, opened 2023-07-08
Token-count integration


jamiemccarthy:
Currently `ruby-openai` has no functionality for dealing with token counts. This is unfortunate because they do play a significant role in OpenAI's API. My app that uses it would benefit from built-in tools that count tokens in an integrated way.

I propose three improvements, which build on each other:

1. Add an `OpenAI::Client#token_count(parameters)` method, which returns two integers: the total number of OpenAI tokens in the parameters' messages, and the max number of tokens (input and output) for the parameter's model. Initially this would simply use [OpenAI's rules of thumb](https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them) to estimate the count, based on character count and word count.
2. As a feature upgrade, `token_count` could check whether the module `::Tiktoken` is defined, and if so use Tiktoken's `.encode.length`, otherwise fall back on the rule-of-thumb estimate. (Since the relevant model is defined in `parameters`, we wouldn't have to guess which encoding to use.)
3. Have `json_post` `raise` a custom exception if the `token_count` for a request does not leave enough tokens for a meaningful response (say, a default of >= 10 tokens). The idea here is to avoid paying to send OpenAI messages it can't fulfill. This would probably include a new initialization option for `OpenAI::Client` to enable this behavior, that initially defaults to not `raise`-ing, but (if y'all find it useful) it might default to `raise`-ing in a future version. Maybe the `parameters` argument could have a new field added that overrides the minimum required token overhead.

I would make use of this functionality in my own app by writing a wrapper around `OpenAI::Client#chat` that instead of one model takes an Array, and tries to `#chat` with each in turn until it finds one that doesn't `raise`. If all 3 of the above were done, the wrapper would be just a couple of lines long. I figure that functionality might be specific enough to myself that it wouldn't be appropriate to add it to the gem, but if y'all are interested, it's possible.

Of course all this can be done manually, but since this gem is targeted at OpenAI, including this functionality seems like a natural fit to me. I've looked for work similar to this but haven't seen any.

marckohlbrugge:
Hitting the token limit is one of the most common issues I run into. Some more functionality around this would be very useful 👍

EthanMarrs:
I would find 1. and 2. useful 😄 I've implemented a very basic rule of thumb calculation for my apps, but a more accurate option would be useful in specific cases.

jamiemccarthy:
[...]

PR #306, opened 2023-08-14
Add token_count client method

jamiemccarthy:
## token_count

As discussed in #291 , it can be useful in many cases to support calculating the token-count of messages before sending them to OpenAI.

This PR adds an `OpenAI::Client.token_count` instance method which does this. To minimize astonishment, it accepts the same `parameters` input as the other methods. Its output is always a nonnegative integer.

If the caller has already `require`'d the `tiktoken_ruby` module, this new method automatically uses it to calculate exactly the number of tokens that OpenAI will parse. (Because `parameters` includes the model name, we know we're using the same algorithm OpenAI will use.)

Or, if `Tiktoken` is not available, it uses the two "helpful rules of thumb" in OpenAI's [docs](https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them) and averages them as an estimate.

Because testing the `Tiktoken` code paths works best if the module is actually included and not mocked, this PR adds that gem to the `Gemfile`. Note that this does not change the dependency list. Also, because the test is written against two specific, existing models, that module's counts should not change in the future and thus the test results should be pretty future-proof.

I'm happy to make any changes desired. Just let me know.

## ruby 2.6

`Tiktoken` requires ruby >= 2.7.  This gem requires ruby >= 2.6. I wanted to keep this gem's test suite passing on all versions of ruby.

What I did was the slightly unorthodox method of editing `Gemfile` to only make mention of `tiktoken` when running on ruby >= 2.7, and add conditionals in the test file to skip those three tests on ruby 2.6.

It works — I checked it locally on ruby 2.6, 2.7, and 3.1.

But... it's a little ugly. Personally, if you asked me, since 2.6 isn't getting security updates, I wouldn't want to encourage its use. I would revert the ugly changes (f8d548103a6f11b8231192894f2951fa3e6de5ef) and set the [required version](https://github.com/alexrudall/ruby-openai/blob/ef982fe99ed3585c2238a33ab891294113fe894a/ruby-openai.gemspec#L12) to 2.7. But I didn't want to make that change without discussing/consulting about it first. What's in this PR works fine and changes things as minimally as possible.

## All Submissions:

### DIFF
--- a/lib/openai/client.rb
+++ b/lib/openai/client.rb
@@ -54,5 +54,32 @@ module OpenAI
     def translate(parameters: {})
       OpenAI::Client.multipart_post(path: "/audio/translations", parameters: parameters)
     end
+
+    def token_count(parameters: {})
+      contents = parameters[:messages]&.map { |message| message[:content] }&.join(" ") || nil
+      return 0 if contents.nil?
+
+      if self.class.tiktoken_defined?
+        model = parameters[:model] || self.class.default_tiktoken_model
+        encoding = Tiktoken.encoding_for_model(model) || Tiktoken.encoding_for_model(self.class.default_tiktoken_model)
+        encoding.encode(contents).size
+      else
+        # https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them
+        count_by_chars = contents.size / 4.0
+        count_by_words = contents.split.size * 4.0 / 3
+        estimate = ((count_by_chars + count_by_words)/2.0).round
+        [1, estimate].max
+      end
+    end
+
+    protected
+
+    def self.tiktoken_defined?
+      !!(defined? ::Tiktoken)
+    end
+
+    def self.default_tiktoken_model
+      "gpt-3.5-turbo".freeze
+    end
   end
 end
--- a/spec/openai/client/token_count_spec.rb
+++ b/spec/openai/client/token_count_spec.rb
@@ -1,0 +1,56 @@
+RSpec.describe OpenAI::Client do
+  describe "#token_count" do
+    let(:messages) { [
+      { content: "Red is my favorite color." },
+      { content: "You miss 100% of the shots you don't take." },
+      { content: "Ingredients for banana bread can include 1 egg or 3 eggs." },
+      { content: "Egg is not a necessary ingredient." }
+    ]}
+    let(:client) { OpenAI::Client.new }
+    let(:model) { nil }
+    let(:token_count) do
+      parameters = { messages: messages }
+      parameters[:model] = model if model
+      client.token_count(parameters: parameters)
+    end
+
+    context "without tiktoken" do
+      context "without model" do
+        it "counts tokens naively" do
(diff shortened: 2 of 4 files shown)


lib/openai/client.rb
```diff
@@ -54,5 +54,32 @@ def transcribe(parameters: {})
     def translate(parameters: {})
       OpenAI::Client.multipart_post(path: "/audio/translations", parameters: parameters)
     end
+
+    def token_count(parameters: {})
```
codenamev, line 58: This feels strange to me living inside the `Client` class.  I'd expect public methods in this context to interface with OpenAI in some way.  Consider moving this to a new utility class or included on the top-level `OpenAI` namespace.

lib/openai/client.rb
```diff
@@ -54,5 +54,32 @@ def transcribe(parameters: {})
...
+      contents = parameters[:messages]&.map { |message| message[:content] }&.join(" ") || nil
```
codenamev, line 59: I'd like to vote against forcing this method to expect multiple messages by default. Consider using a single `content = ""` parameter and reducing the complexity here.

alexrudall: Agree with this

spec/openai/client/token_count_spec.rb
```diff
@@ -0,0 +1,60 @@
+RSpec.describe OpenAI::Client do
+  describe "#token_count" do
+    let(:messages) { [
+      { content: "Red is my favorite color." },
+      { content: "You miss 100% of the shots you don't take." },
+      { content: "Ingredients for banana bread can include 1 egg or 3 eggs." },
+      { content: "Egg is not a necessary ingredient." }
```
alexrudall, line 7: 🥚

jamiemccarthy:
> * Rename the method to `rough_token_count`

I'm happy to do this, but I'd find it less useful. Does it change your mind to know that OpenAI's recommended estimation algorithms are usually high by 10-20%?

To me, that's a lot of wiggle room for error, if people are trying to have their code make decisions based on that estimate.

For example, in the largest example given in [the docs](https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them), the Declaration of Independence, it's off from Tiktoken's calculations by 13-17%:

```
jamie@Jamies-MacBook-Air ruby-openai % bundle exec irb                                         
irb(main):001:0> require 'openai'
=> true
irb(main):002:0> OpenAI::Client.new.token_count(parameters: { model: "tktk", messages: [ { content: File.read("dofi")}]})
=> 1887
irb(main):003:0> require 'tiktoken_ruby'
=> true
irb(main):004:0> Tiktoken.encoding_for_model("gpt-4").encode(File.read("dofi")).size
=> 1616
irb(main):005:0> Tiktoken.encoding_for_model("davinci").encode(File.read("dofi")).size
=> 1674
```

jamiemccarthy:
Thanks for your feedback. I've submitted #312 which I believe addresses it.

PR #312, opened 2023-08-18
Add OpenAI.rough_token_count class method

jamiemccarthy:
It can be useful to estimate the token count of some text before sending it to OpenAI. This PR adds an `OpenAI.rough_token_count` class method which does this.

To perform the estimation, it uses the two "helpful rules of thumb" in OpenAI's [docs](https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them) and takes the arithmetic average.

The empty string returns 0, and any other string returns an integer greater than 0.

This is a rewrite of #306 to remove the Tiktoken functionality and change the interface as requested.

## All Submissions:

### DIFF
--- a/lib/openai.rb
+++ b/lib/openai.rb
@@ -53,6 +53,8 @@ module OpenAI
     yield(configuration)
   end
 
+  # Estimate the number of tokens in a string, using the rules of thumb from OpenAI:
+  # https://help.openai.com/en/articles/4936856-what-are-tokens-and-how-to-count-them
   def self.rough_token_count(content = "")
     raise ArgumentError, "rough_token_count requires a string" unless content.is_a? String
     return 0 if content.empty?
--- a/lib/openai.rb
+++ b/lib/openai.rb
@@ -52,4 +52,14 @@ module OpenAI
   def self.configure
     yield(configuration)
   end
+
+  def self.rough_token_count(content = "")
+    raise ArgumentError, "rough_token_count requires a string" unless content.is_a? String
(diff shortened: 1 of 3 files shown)


lib/openai.rb
```diff
@@ -52,4 +52,14 @@ def self.configuration
   def self.configure
     yield(configuration)
   end
+
+  def self.rough_token_count(content = "")
```
alexrudall, line 56: Can you add your explanation from the PR description here

```suggestion

  # Explain what this does
  # Link to OpenAI docs
  def self.rough_token_count(content = "")
```

alexrudall (changes requested):
This is great, thanks a lot for rewriting. Just some documentation needed, would you mind adding a section to the README with a couple lines briefly explaining how to estimate tokens in a string and link to the OpenAI docs. Probably between Azure and Models sections I think? 

And couple inline comments also, just with the method.

lib/openai.rb
jamiemccarthy: Done, in 3845a45a8664b6b0c5f0096d303622ca017e58bd.

If [that new section in the README](https://github.com/alexrudall/ruby-openai/blob/3845a45a8664b6b0c5f0096d303622ca017e58bd/README.md#counting-tokens) is too wordy, let me know and I'll cut it down.

alexrudall:
This is perfect thanks @jamiemccarthy!

alexrudall (approved):

PR #323, opened 2023-09-01
:memo: fix an small error in the README.md documentation

tonystrawberry:
## All Submissions:

## Description

- Calling `client.files.upload` already returns the response as a Hash so we should not call `.body` on the response.


alexrudall (approved):
Good spot, thank you!

Issue #255, opened 2023-05-02
Accessing to response headers and code in v4


dmitry:
**Is your feature request related to a problem? Please describe.**
In version 3 there was a possibility to access the response code. In the current version 4.0.0, it's not possible.

**Describe the solution you'd like**
It would be good to access the response headers and code. Sometimes openai API returns 429 response code. Need to be able to retry in 1 second once again. Now it's not possible to use a retry mechanism.

**Describe alternatives you've considered**
Stick with v3 for now, as v4 returns via `to_json` JSON hash.

airblade:
The API uses [HTTP error codes](https://platform.openai.com/docs/guides/error-codes) with its responses to convey useful information.  It would be very helpful to be able to access them.

alexrudall:
Good suggestion. We could add the option to pass `raw: true` or similar and that will return the raw responses instead. Would that work?

airblade:
Sounds good to me.

ScotterC:
+1 to this. It's particularly important when writing logic to handle retries and stay under rate limiting.

It looks like in [http](https://github.com/alexrudall/ruby-openai/blob/main/lib/openai/http.rb) that each request method is returning `body` if it's there. I'm guessing that passing each of these through a general `build_response` method to create a response object would be appropriate?

kevingriffin:
It'd be great if error statuses actually resulted in exceptions being raised. Getting the raw response could be helpful sometimes, but ideally the error cases wouldn't seem to be normal responses until you went to parse them.

thibaut-pro:
I'd love to upgrade to v5 to use the `rough_token_count` but this issue is a blocker for us.

> the option to pass `raw: true` or similar and that will return the raw responses instead.

☝️ That would work really well!

ScotterC:
@alexrudall heres my solution to this that I've been running in production for 3+ months. If it looks good I can setup a PR. 
https://github.com/ScotterC/ruby-openai/blob/dewey/lib/openai/http.rb

`raw` is a Boolean parameter passed into json_post with other parameters. It determines whether to return the raw faraday response or not.

cheiff:
Yes, please submit a PR with your patch. I've tested your code and works OK. This is what we need, too.
Thank you!

codenamev:
> It'd be great if error statuses actually resulted in exceptions being raised. Getting the raw response could be helpful sometimes, but ideally the error cases wouldn't seem to be normal responses until you went to parse them.

[...]

PR #327, opened 2023-09-05
Adding `raw` as optional parameter to expose faraday object

ScotterC:
Faraday object exposes status, body and headers

Explanation
- Currently json_post only gives you the body as json and not access to the response object.
- `raw` parameter is not passed to LLM provider
- `raw` parameter determines whether to return Faraday object and convert only the body to json

Resolves https://github.com/alexrudall/ruby-openai/issues/255

## All Submissions:

### DIFF
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -17,7 +17,8 @@ module OpenAI
         end
 
         req.headers = headers
-        req.body = parameters.except(:raw).to_json
+        new_params = parameters.reject { |key, _| key == :raw }
+        req.body = new_params.to_json
       end
 
       return nil unless response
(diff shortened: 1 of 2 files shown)


ScotterC:
@alexrudall passing rubocop requires breaking down the json_post method which is a much larger PR to parse. If the direction looks good then I'll go about it.

alexrudall:
The latest version [v6](https://github.com/alexrudall/ruby-openai/releases/tag/v6.0.0) raises HTTP errors as exceptions - I believe catching these errors was the main use-case for this. Closing with that in mind - we can reopen if there is still a usecase for this

PR #328, opened 2023-09-07
Adds ability to capture HTTP errors

### LINKED ISSUE #255: Accessing to response headers and code in v4

codenamev:
As of v4.0.0 of this gem, we lost the ability to access response headers and response status code.  Access to these is critical in safely handling HTTP failures that can happen when reaching out to OpenAI's API.

This introduces a new configuration option `raise_error` that allows users to raise a new `OpenAI::HTTP::Error` anytime the response status code is 4xx or 5xx, and provides access to the response's `status`, `headers`, and `body` on the error instance that is raised.

In #327, an approach to resolve this allows forcing a "raw" response to be returned.  For the specific case of handling HTTP errors though, this is not a straightforward solution that I would expect as a first-time user of the library.

Since Faraday is used, they have an excellent [`RaiseError`](https://github.com/lostisland/faraday/blob/v1.0.0/lib/faraday/response/raise_error.rb#L7) middleware ([recommended in their docs](https://lostisland.github.io/faraday/#/middleware/included/raising-errors?id=raising-errors)) that will raise a [Faraday::Error](https://github.com/lostisland/faraday/blob/v1.0.0/lib/faraday/error.rb) for any 4xx or 5xx response status codes.  Instances of this error contain a [`response` Hash attribute](https://github.com/lostisland/faraday/blob/v1.0.0/lib/faraday/response/raise_error.rb#L41) that include the `status`, `headers`, and `body` of the response.  This takes advantage of that error class and wraps it in a library-specific one so that end-users can rescue from this in a convenient way.

Some dis-advantages to this solution is that it is coupled with Faraday.  However, extending the new `OpenAI::HTTP::Error` class to conform to the same return object for the `response` accessor (a `Hash` with `status`, `headers`, and `body` keys) should be fairly straightforward if it is decided to move away from Farday in the future.  I have tried my best to avoid any Farday-specific mentions in the README for that reason.

Fixes #255

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,59 @@ OpenAI.configure do |config|
 end
 ```
 
+#### Handling Errors from the OpenAI API
+
+The default, requests using this library will only return the `body` of the response (if one exists).  You can change this behavior and have 4xx and 5xx HTTP status codes raise an `OpenAI::HTTP::Error` by using the following client configuration:
+
+```ruby
+client = OpenAI::Client.new(
+    access_token: "access_token_goes_here",
+    raise_error: true
+)
+```
+
+or when configuring the gem:
+
+```ruby
+OpenAI.configure do |config|
+    config.access_token = ENV.fetch("OPENAI_ACCESS_TOKEN")
+    config.raise_error = true
+end
+```
+
+With this configuration set, calls to any of the client methods (`chat`,
+`completions`, `embeddings`, etc.) will all raise an error
+(`OpenAI::HTTP::Error`) if the response's HTTP status code is 4xx or 5xx. This
+special error gives you access to some key information about the response on the instance of the error raised:
+
+```ruby
+client = OpenAI::Client.new(
+    access_token: "access_token_goes_here",
+    raise_error: true
+)
+begin
+  response = client.chat(
+      parameters: {
+          model: "gpt-42", # providing an invalid model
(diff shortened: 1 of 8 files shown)


codenamev:
@alexrudall I left the rubocop failures since resolving them will require some re-organization of the `HTTP` module (due to it's line-length).  If this idea is acceptable, I can circle back and adjust; but I'd love to just bump that number in the rubocop config since it's only off by a few lines.

atesgoral:
@codenamev In #342 I'm taking a more opinionated approach of not making error throwing optional. Also found a way throw the Faraday exceptions while streaming.

alexrudall:
Really appreciate this contribution @codenamev - going with #344 as it's a little simpler and includes error details by default. BUT contributions like this help explore alternatives so it's seriously appreciated. 🙏

PR #332, opened 2023-09-27
Better SSE parsing

atesgoral:
The issue mentioned in #251 persists.

OpenAI reliably returns one completion JSON per SSE message. However, when intermediaries are placed between OpenAI and the client (e.g. a proxy), the combining and splitting of chunks during streaming/buffering (or even TCP fragmentation) can cause the client to receive chunks with completion fragments (invalid JSON).

This PR brings more [spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation)-compliant SSE parsing. Instead of stripping and ignoring whitespace, processes a chunk by splitting it into blocks delimited by `"\n\n"`, and buffering any remaining block that is not a full `data` field.

**More** spec-compliant, but not really fully spec-compliant: There's still no handling of multi-line data spread across adjacent `data` fields. But there's no need to make this a generic SSE parser since the only concern is hitting OpenAI.

I've updated the http tests by making them emit proper SSE chunks into the stream. I added a test for the fragmented completion JSON case.

Note that this is still a bit of a quick hack and not the best way to parse SSE. An ideal parser would process the stream line by line with a state machine (instead of YOLO splitting at `\n\n`).

I also noticed two issues that I did not address in the scope of this PR:

1. I don't think there's such thing as an `error` field in SSE. It's neither in the spec, nor in OpenAI's documentation. It would be nice to remove the handling of that, or mention in a code comment where it's coming from.
2. The code simply ignores JSON parse errors. IMHO, the library instead should be unforgiving and error out when an unexpected payload is received. The ignoring of parse errors is a convenience for not directly handling the `"[DONE]"` value omitted by OpenAI, and also the very JSON fragmentation issue that is not very common, but still a possibility in certain settings.

### DIFF
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -1,5 +1,9 @@
+require_relative "sse"
+
 module OpenAI
   module HTTP
+    include OpenAI::SSE
+
     def get(path:)
       to_json(conn.get(uri(path: path)) do |req|
         req.headers = headers
@@ -53,37 +57,13 @@ module OpenAI
     # @param user_proc [Proc] The inner proc to call for each JSON object in the chunk.
     # @return [Proc] An outer proc that iterates over a raw stream, converting it to JSON.
     def to_json_stream(user_proc:)
-      @buffer = ""
-
-      proc do |chunk, _|
-        chunk = @buffer + chunk
-        @buffer = ""
-
-        blocks = chunk.split("\n\n", -1)
-
-        while blocks.length.positive?
-          block = blocks.shift
-
-          if blocks.empty?
-            @buffer = block unless block.empty?
-            break
-          end
-
-          matches = /([^:]+):\s*(.+)/.match(block)
-
-          next unless matches
-
-          field_name, field_value = matches.captures
-
-          next unless %w[data error].include?(field_name)
-
-          begin
-            user_proc.call(JSON.parse(field_value))
-          rescue JSON::ParserError
-            # Ignore invalid JSON.
-          end
-        end
+      completion_json_proc = proc do |completion_json|
+        user_proc.call(JSON.parse(completion_json))
+      rescue JSON::ParserError
+        # Ignore invalid JSON.
       end
+
+      to_completion_json_stream(completion_json_proc: completion_json_proc)
     end
 
     def conn(multipart: false)
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -53,11 +53,35 @@ module OpenAI
     # @param user_proc [Proc] The inner proc to call for each JSON object in the chunk.
     # @return [Proc] An outer proc that iterates over a raw stream, converting it to JSON.
     def to_json_stream(user_proc:)
+      @buffer = ""
+
       proc do |chunk, _|
-        chunk.scan(/(?:data|error): (\{.*\})/i).flatten.each do |data|
-          user_proc.call(JSON.parse(data))
-        rescue JSON::ParserError
-          # Ignore invalid JSON.
+        chunk = @buffer + chunk
+        @buffer = ""
+
+        blocks = chunk.split("\n\n", -1)
+
+        while blocks.length.positive?
+          block = blocks.shift
+
+          if blocks.empty?
+            @buffer = block unless block.empty?
+            break
+          end
+
+          matches = /([^:]+):\s*(.+)/.match(block)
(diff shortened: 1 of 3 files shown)


lib/openai/http.rb
```diff
@@ -53,13 +57,13 @@ def to_json(string)
     # @param user_proc [Proc] The inner proc to call for each JSON object in the chunk.
     # @return [Proc] An outer proc that iterates over a raw stream, converting it to JSON.
     def to_json_stream(user_proc:)
```
atesgoral, line 59: A better name for this would be `to_completion_stream` because once the completion JSON is parsed, it's not longer "JSON" -- it's a completion object.

lib/openai/sse.rb
atesgoral, line 1: I didn't add a separate unit test for this new module because it's sufficiently tested by the http tests.

spec/openai/client/http_spec.rb
```diff
@@ -108,7 +108,11 @@
       context "when called with a string containing a single JSON object" do
         it "calls the user proc with the data parsed as JSON" do
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
-          stream.call('data: { "foo": "bar" }')
+          stream.call(<<~CHUNK)
+            data: { "foo": "bar" }
+
+            #
```
atesgoral, line 114: This ensures the trailing `\n\n` are preserved.

spec/openai/client/http_spec.rb
```diff
@@ -117,13 +121,14 @@
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
           expect(user_proc).to receive(:call).with(JSON.parse('{"baz": "qud"}'))
 
-          stream.call(<<-CHUNK)
+          stream.call(<<~CHUNK)
```
atesgoral, line 124: Should be unindenting these, so switched from `-` to `~` in these heredocs.

spec/openai/client/http_spec.rb
```diff
@@ -134,44 +139,52 @@
...
+            stream.call("#{chunk}\n\n")
           end
         end
       end
 
       context "when called with a string containing that looks like a JSON object but is invalid" do
-        let(:chunk) do
-          <<-CHUNK
-            data: { "foo": "bar" }
-            data: { BAD ]:-> JSON }
```
atesgoral, line 146: So, this was actually a totally new case that doesn't have to be handled in the scope of OpenAI responses because they don't seem to return anything like this. Normally, adjacent `data` fields that the values should be concatenated with a `\n` -- it's a multi-line value. In this case, the above would be read as `"{ "foo": "bar" }\n{ BAD ]:-> JSON }"` and that is not what this test's intention is.

spec/openai/client/http_spec.rb
```diff
@@ -134,44 +139,52 @@
...
           expect(user_proc).to receive(:call).with(
             JSON.parse('{ "message": "A bad thing has happened!" }')
           )
 
           expect do
-            stream.call(chunk)
+            stream.call(<<~CHUNK)
+              data: { "foo": "bar" }
+
+              error: { "message": "A bad thing has happened!" }
```
atesgoral, line 174: AFAIK, there's no such thing as an "error" field. But that is beyond the scope of this PR.

spec/openai/client/http_spec.rb
```diff
@@ -134,44 +139,52 @@
...
+
+              #
+            CHUNK
+          end.not_to raise_error
+        end
+      end
+
+      context "when called with JSON split across chunks" do
```
atesgoral, line 182: This is the not-so-common case this PR intends to fix. Most people using this library to hit OpenAI directly will not encounter this.

chitalian:
This is awesome! Thanks for working on this @atesgoral

atesgoral:
Related issue: Helicone/helicone/issues/678
Example of some optimistic client not being able to handle fragmentation. And Helicone's workaround for working around those clients by doing some internal buffering: https://github.com/Helicone/helicone/blob/1e3bb5d2c2cac592f3ad2a7812bede86e24699bb/worker/src/lib/HeliconeProxyRequest/handler.ts#L29-L51

atesgoral:
I'm working on a proper parser. I'll report back.

PR #336, opened 2023-10-04
Robust SSE parsing

burke:
I took @atesgoral's work in #332 and ran with it a bit

A few things fixed up or adjusted here:

1. OpenAI always sends "data: [DONE]\n\n" at the end of responses; we can assert that was received to detect aborted connections.
2. OpenAI never sends "error: ...\n\n" messages: errors are passed inside the JSON object attached to the "data: {...}\n\n".
3. I'm not aware of OpenAI ever sending malformed JSON, so erroring on that (new behaviour) feels correct to me.

I also made a few other changes along the way:

1. Mostly switched from require_relative to require
2. Changed up-front nested class/module loading to autoloads for (trivial runtime performance gain for consumers and) making it possible to refer to OpenAI::Error in OpenAI::SSE without having to put the requires at the bottom of openai.rb.

Let me know what you think, I'm happy to keep adjusting here.

## All Submissions:


vizakenjack:
There is an issue that should be fixed (both in original ruby-openai and your pull request).

How to reproduce: try to send a pretty long text that won't fit into context. There will be a message:

atesgoral:
@vizakenjack That error sounds like just hitting token limits. What was your fix?

fabioxgn (approved):
@alexrudall can we get this merged? I'm having an issue where the chunks are being split in half and the regex on `to_json_stream` does not match the partial chunk, so it's ignored.

This branch fixes that issue.

This is what I'm getting in `to_json_stream`:

```
chunk
=> "98,\"model\":\"gpt-4-0613\",\"choices\":[{\"index\":0,\"delta\":{\"role\":\"assistant\",\"content\":\"\"},\"finish_reason\":null}]}\n\ndata: {\"id\":\"chatcmpl-8FTgAGg922FFFVvUVdLp8raDbny4j\",\"object\":\"chat.completion.chunk\",\"crea"
```

This does not match `chunk.scan(/(?:data|error): (\{.*\})/i)` so this chunk is ignored.

alexrudall:
Hey @fabioxgn, have you tried v5.2? That should fix your issue. (I still
probably will merge this)

alexrudall:
Big thanks for your work on this @burke - SSE parsing is now much improved / fixed with @atesgoral's PR, so I think this one is not needed. Definitely open to the autoload changes if you want to cherrypick them to a new PR. Thanks!

PR #338, opened 2023-10-16
Proper SSE parsing

atesgoral:
Follow up to #332

1. Switch to using `event_stream_parser`, a spec-compliant event stream parser (that I recently published)
2. ~~The stream proc now receives an optional second argument for any errors encountered during parsing -- consumers should know, instead of these errors being silently ignored by this library~~
3. Explicitly check for `"[DONE]"` chunks.
4. ~~Remove bits about errors OpenAI does not send~~
5. ~~Tweak bit about token usage in streams (it's not a public feature yet)~~
6. Check HTTP response code and only try to parse the body as an error JSON if it's not OK.

## All Submissions:

### DIFF
--- a/spec/openai/client/http_spec.rb
+++ b/spec/openai/client/http_spec.rb
@@ -108,42 +108,39 @@ RSpec.describe OpenAI::HTTP do
       context "when called with a string containing a single JSON object" do
         it "calls the user proc with the data parsed as JSON" do
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
-          stream.call('data: { "foo": "bar" }')
+          stream.call(<<~CHUNK)
+            data: { "foo": "bar" }
+
+            #
+          CHUNK
         end
       end
 
-      context "when called with string containing more than one JSON object" do
+      context "when called with a string containing more than one JSON object" do
         it "calls the user proc for each data parsed as JSON" do
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
           expect(user_proc).to receive(:call).with(JSON.parse('{"baz": "qud"}'))
 
-          stream.call(<<-CHUNK)
+          stream.call(<<~CHUNK)
             data: { "foo": "bar" }
 
             data: { "baz": "qud" }
 
             data: [DONE]
 
+            #
           CHUNK
         end
       end
 
-      context "when called with a string that does not even resemble a JSON object" do
-        let(:bad_examples) { ["", "foo", "data: ", "data: foo"] }
-
-        it "does not call the user proc" do
-          bad_examples.each do |chunk|
-            expect(user_proc).to_not receive(:call)
-            stream.call(chunk)
-          end
-        end
-      end
-
-      context "when called with a string containing that looks like a JSON object but is invalid" do
+      context "when called with string containing invalid JSON" do
         let(:chunk) do
-          <<-CHUNK
+          <<~CHUNK
             data: { "foo": "bar" }
-            data: { BAD ]:-> JSON }
+
+            data: NOT JSON
+
+            #
           CHUNK
         end
 
@@ -156,22 +153,31 @@ RSpec.describe OpenAI::HTTP do
         end
       end
 
-      context "when called with a string containing an error" do
-        let(:chunk) do
-          <<-CHUNK
-            data: { "foo": "bar" }
-            error: { "message": "A bad thing has happened!" }
-          CHUNK
-        end
+      context "when OpenAI returns an HTTP error" do
+        let(:chunk) { "{\"error\":{\"message\":\"A bad thing has happened!\"}}" }
+        let(:env) { Faraday::Env.new(status: 500) }
 
-        it "does not raise an error" do
-          expect(user_proc).to receive(:call).with(JSON.parse('{ "foo": "bar" }'))
+        it "does not raise an error and calls the user proc with the error parsed as JSON" do
           expect(user_proc).to receive(:call).with(
-            JSON.parse('{ "message": "A bad thing has happened!" }')
+            {
+              "error" => {
+                "message" => "A bad thing has happened!"
+              }
+            }
           )
 
           expect do
-            stream.call(chunk)
+            stream.call(chunk, 0, env)
+          end.not_to raise_error
+        end
+      end
+
+      context "when called with JSON split across chunks" do
+        it "calls the user proc with the data parsed as JSON" do
+          expect(user_proc).to receive(:call).with(JSON.parse('{ "foo": "bar" }'))
+          expect do
+            stream.call("data: { \"foo\":")
+            stream.call(" \"bar\" }\n\n")
           end.not_to raise_error
         end
       end
--- a/spec/openai/client/http_spec.rb
+++ b/spec/openai/client/http_spec.rb
@@ -153,6 +153,26 @@ RSpec.describe OpenAI::HTTP do
         end
       end
 
+      context "wehn called with string containing Obie's invalid JSON" do
+        let(:chunk) do
+          <<~CHUNK
+            data: { "foo": "bar" }
+
+            data: {"id":"chatcmpl-123","object":"chat.completion.chunk","created":123,"model":"gpt-4","choices":[{"index":0,"delta":{"role":"assistant","content":null,"fu
+
+            #
+          CHUNK
+        end
+
+        it "does not raise an error" do
+          expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
+
+          expect do
+            stream.call(chunk)
+          end.not_to raise_error
+        end
+      end
+
       context "when OpenAI returns an HTTP error" do
         let(:chunk) { "{\"error\":{\"message\":\"A bad thing has happened!\"}}" }
         let(:env) { Faraday::Env.new(status: 500) }
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     ruby-openai (5.1.0)
+      event_stream_parser (~> 0.3.0)
       faraday (>= 1)
       faraday-multipart (>= 1)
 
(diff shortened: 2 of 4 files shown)


ScotterC:
Thanks for this. Can you link the EventStreamParser gem? I wasn't able to see it in your github but I read through it locally. I've merged this into my fork and trying it out in production.

atesgoral:
@ScotterC It's publicly published: https://rubygems.org/gems/event_stream_parser

atesgoral:
@andresgutgon Ah, my apologies. I forgot about making the repo public. It is now :)

smojtabai:
Hey @atesgoral , I was testing this, but I think it still swallows errors, not sure if this is a spec issue with OpenAI or what:

OpenAI::Client.new.chat(
      parameters: {
        model: "gpt-4",
        messages: [{ role: "garbage", content: "hello"}],
        temperature: 0,
        stream: proc do |chunk, error|
          puts "inside stream"
          puts chunk
          puts error
        end
      }
    )

It looks like openAI returns an entire error json object and this https://github.com/Shopify/event_stream_parser/blob/main/lib/event_stream_parser.rb skips it?

atesgoral:
@smojtabai `event_stream_parser` is a pure parser of event streams and is agnostic of OpenAI. It's not swallowing any events.

It seems ruby-openai isn't checking the HTTP response code before processing the response as an event stream. In case of errors, OpenAI returns a JSON response, not event stream. I just realized this and I'll see if I can forward-fix that in this PR. Thanks for raising it.

smojtabai:
@atesgoral ah yes, you are correct, I miss understood what was going on there. 

https://github.com/alexrudall/ruby-openai/pull/328 looks like tried to contribute a fix but was never merged or looked at, unsure if @alexrudall has any thoughts or if it will be merged ?

atesgoral:
With a one-line additional here: https://github.com/alexrudall/ruby-openai/blob/e76c729e0bf7ab32d66e52bbb0d49baebe372fdf/lib/openai/http.rb#L66-L69

```rb
f.response :raise_error
```

It is possible to mount the [RaiseError middleware](https://lostisland.github.io/faraday/#/middleware/included/raising-errors) to get 4xx and 5xx responses percolate all the way up, but it's better to catch 4xx errors locally to be able to grab the JSON OpenAI is returning in the body.

I'll take a look at #328 to see if I can find inspiration there.

andresgutgon:
This is looking fantastic. Looking forward to see this finish. Great work @atesgoral 👏

atesgoral:
Complication: When the `RaiseError` middleware is enabled, Faraday will bail out (like it should) early and not call the `on_data` block. So, in streaming mode, there's no way to access the response body in the presence of that stock middleware.

Ignoring the middleware, I couldn't get access to Faraday's `env` argument in the `on_data` block for some reason (maybe it's not in the version of Faraday that this gem relies on?). If I could, then I could check the HTTP response code within the `on_data` block while also collecting response chunks.

Contemplating on creating a streaming-friendly version of the RaiseError middleware (and/or upgrading Faraday if needed).

smojtabai:
Hmmm, I think upgrading Faraday would be the best looking solution. As a hack I tried monkey patching the http json_pst method with this:

`def json_post(path:, parameters:)
      stream_body = ''
      is_streaming = parameters[:stream].respond_to?(:call)
      response = conn.post(uri(path: path)) do |req|
        if is_streaming
          to_json_stream_proc = to_json_stream(user_proc: parameters[:stream])
          req.options.on_data = proc do |chunk, chunk_size|
            stream_body = chunk
            to_json_stream_proc.call(chunk, chunk_size)
          end

          parameters[:stream] = true # Necessary to tell OpenAI to stream.
        elsif parameters[:stream]
          raise ArgumentError, "The stream parameter must be a Proc or have a #call method"
        end

        req.headers = headers
        req.body = parameters.to_json

      end
      if is_streaming && response.status >= 400
        to_json(stream_body)
      else
        to_json(response&.body)
      end
    end`

It sorta for now keeps the same signature on return of an error, I guess we could even call the user function instead this way if needed. Unsure if this can cause any issues with race conditions or if its a good idea but throwing it out there.

If we don't want to check the return code, we could just augment the to_json_stream function as follows which is hackier but may have less side effects:

`def to_json_stream(user_proc:)
      parser = EventStreamParser::Parser.new

      proc do |chunk, chunk_size|
        processed_data = false
        parser.feed(chunk) do |type, data, id, reconnection_time|
          processed_data = true
          parsed_data = JSON.parse(data)
          user_proc.call(parsed_data) unless data == "[DONE]"
        end
        if processed_data == false
          Rails.logger.warn("Failed to parse chunk: #{chunk}")
          parsed_data = JSON.parse(chunk)
          if parsed_data['error'].present?
            raise StandardError.new("Error with OpenAI: #{parsed_data}")
          end
        end
      end
    end`

This basically checks if the stream processor fails it tries to see if JSON got returned and for now throws (but can do what we want) in case of error.

smojtabai:
We also may have wondered a bit and I apologize, I do think your first fix is important and can be merged separate from this but streaming in my opinion is not very usable without the ability to understand error responses from OpenAI. Do you have any connection with @alexrudall , I notice he hasn't commented on the other PR's trying to fix the error issue?

atesgoral:
Here's how I managed to grab the error response and raise the appropriate Faraday exception with it:

https://github.com/alexrudall/ruby-openai/commit/13ef7e34bc55123898e94dcf98b55074d17b8d77

```rb
      parser = EventStreamParser::Parser.new
      proc do |chunk, _bytes, env|
        if env && env.status != 200
          raise_error = Faraday::Response::RaiseError.new
          raise_error.on_complete(env.merge(body: JSON.parse(chunk)))
        end

        parser.feed(chunk) do |_type, data|
          user_proc.call(JSON.parse(data)) unless data == "[DONE]"
        end
      end
    end
```

This also feels a bit like a Faraday deficiency of not making this super easy.

atesgoral:
@smojtabai No, Alex and I are not connected. I'm just trying to fix things for our project (on my fork of this library) while trying to contribute back to the source.

atesgoral:
So, I might actually tweak this PR to preserve existing functionality of treating HTTP errors as non-errors and returning the JSON error as a result. #342 could be a bigger leap to change behaviour, as a follow-up.

smojtabai:
I agree with keeping existing functionality for now, more likely to get merged, thank you!

spec/openai/client/http_spec.rb
```diff
@@ -108,42 +108,39 @@
...
-        it "does not call the user proc" do
-          bad_examples.each do |chunk|
-            expect(user_proc).to_not receive(:call)
-            stream.call(chunk)
-          end
-        end
-      end
-
-      context "when called with a string containing that looks like a JSON object but is invalid" do
+      context "when called with string containing invalid JSON" do
```
atesgoral, line 136: It was weird to test for "semblance" of JSON. It's either valid JSON or not.

spec/openai/client/http_spec.rb
```diff
@@ -156,22 +153,31 @@
...
           )
 
           expect do
-            stream.call(chunk)
+            stream.call(chunk, 0, env)
+          end.not_to raise_error
+        end
+      end
+
+      context "when called with JSON split across chunks" do
```
atesgoral, line 175: This is the meat of this PR. ruby-openai now becomes resilient to packet fragmentation at awkward locations.

atesgoral:
Alright, I update the PR to be laser-focused on solving the parsing issue, without touching any other behaviour.

spec/openai/client/http_spec.rb
```diff
@@ -108,42 +108,39 @@
       context "when called with a string containing a single JSON object" do
         it "calls the user proc with the data parsed as JSON" do
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
-          stream.call('data: { "foo": "bar" }')
+          stream.call(<<~CHUNK)
```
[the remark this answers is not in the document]
andresgutgon: Already learning new things keep going 💪

atesgoral:
Follow-up PR to raise errors: #344

smojtabai:
@atesgoral if this branch were merged, would I be able to tell if an error occurred? Did you feel like the callback having an error argument was too big a change (without raising an exception)?

atesgoral:
@smojtabai With this change, there's no change to how errors are subtly handled. You'd still have to check if the parsed JSON chunk your proc receives looks like an error (has an `error` property).

Ah, I guess the shape of the error is changing now since the error value is not being plucked. 🤔

But, assuming the error raising PR could hopefully be also reviewed and accepted in quick succession, there could be a major bump to the gem for the new error handling flavour. We'll see soon 🤞

alexrudall:
Released in [5.2.0](https://rubygems.org/gems/ruby-openai/versions/5.2.0) - huge thanks to you @atesgoral for your work on this and to others for input

spec/openai/client/http_spec.rb
athyuttamre: Hi @atesgoral, I'm an engineer at OpenAI investigating this response. Do you have any logs (ideally from `curl` showing the full response stream) to demonstrate this issue? Please feel free to email me at atty@openai.com. Thanks!

atesgoral: @athyuttamre I haven't recently seen this happen when directly talking to the OpenAI client, but I recall seeing it a while back, when network conditions were bad. But it can still happen when a proxy or some other network element buffers and chops up the text/event-stream chunks at non-event-stream boundaries.

atesgoral: I don't think this is an OpenAI problem at all. Clients just need to be resilient and do buffered/proper parsing of event streams.

atesgoral: For posterity, here are two chunks (stream fragments) from an earlier reproduction of the parser issue that gets unearthed by buffering + rechunking:

Chunk 1:

```
data: {"id":"chatcmpl-83QVo11UROyI8HUeixAo25Vjx0SA5","object":"chat.completion.chunk","created":1695827588,"model":"gpt-4-0613","choices":[{"index":0,"delta":{"role":"assistant","content":""},"finish_reason":null}]}
data: {"id":"chatcmpl-83QVo11UROyI8HUeixAo25Vjx0SA5","object":"chat.completion.chunk","created":1695827588,"model":"gpt-4-0613","choices":[{"index":0,"delta"
```

Chunk 2:

```
:{"content":"Hello"},"finish_reason":null}]}
```

PR #343, opened 2023-10-19
Fix streaming data parsing in OpenAI::HTTP's to_json_stream

mrbokx:
Problem:
The to_json_stream method failed to handle JSON objects split across multiple data packets from the Azure AI service, leading to JSON parsing errors and incomplete/missing data.

Solution:
Implemented a buffering mechanism within to_json_stream to accumulate data until a complete JSON object is formed. This ensures proper parsing and data integrity, regardless of how the data packets are received.

### DIFF
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -54,10 +54,16 @@ module OpenAI
     # @return [Proc] An outer proc that iterates over a raw stream, converting it to JSON.
     def to_json_stream(user_proc:)
       proc do |chunk, _|
-        chunk.scan(/(?:data|error): (\{.*\})/i).flatten.each do |data|
-          user_proc.call(JSON.parse(data))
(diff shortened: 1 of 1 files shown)


mrbokx:
There are a few RuboCop errors present, but they're unrelated to the changes I've introduced with my fix. I've decided to keep these outside the scope.

atesgoral:
Hey @mrbokx, simple buffering won't cut it. Even though event streams look trivial to parse with simple regex + string manipulation, a proper state machine needs to be involved to handle all edge cases.

I'm fixing it here: #338

alexrudall:
@atesgoral 's fix released in [5.2.0](https://rubygems.org/gems/ruby-openai/versions/5.2.0) - thank you for your work on this @mrbokx !

PR #344, opened 2023-10-19
Raise exceptions

atesgoral:
> **Note**
> This PR builds on #338. ~~I unfortunately can't set the base when cutting a PR from a forked repo. Reviewing it commit by commit after skipping the first commit would be helpful.~~
> 
> - [x] #338

- **BREAKING CHANGE**: ruby-openai now raises errors instead of silently ignoring them. It is better for production apps to fail fast and take note of any errors instead of ignoring them. OpenAI is not expected to return any invalid payload in its normal operation.
- Removed the completion test case from the finetunes suite because it wasn't adding any value since it doesn't really exercise an actual fine-tuned model. Besides, the create cassette was broken (had 400 responses throughout). The reason that test was passing is because ruby-openai was eating those errors.
- Tweaks to README: 
  - Tweak/add some bits about token counting during streaming
  - Replace "ChatGPT" with "GPT" (the former is a web app, the latter is the actual model)

## All Submissions:


spec/openai/client/http_spec.rb
```diff
@@ -144,32 +144,12 @@
...
+        it "raise an error" do
           expect(user_proc).to receive(:call).with(JSON.parse('{"foo": "bar"}'))
 
           expect do
             stream.call(chunk)
-          end.not_to raise_error
-        end
-      end
-
-      context "wehn called with string containing Obie's invalid JSON" do
```
atesgoral, line 156: The existing invalid JSON test is enough with this change

alexrudall:
Thanks @atesgoral - taking a look now

README.md
```diff
@@ -161,9 +161,9 @@ client.models.retrieve(id: "text-ada-001")
   - text-babbage-001
   - text-curie-001
 
-### ChatGPT
+### GPT
 
-ChatGPT is a model that can be used to generate text in a conversational style. You can use it to [generate a response](https://platform.openai.com/docs/api-reference/chat/create) to a sequence of [messages](https://platform.openai.com/docs/guides/chat/introduction):
+GPT is a model that can be used to generate text in a conversational style. You can use it to [generate a response](https://platform.openai.com/docs/api-reference/chat/create) to a sequence of [messages](https://platform.openai.com/docs/guides/chat/introduction):
```
alexrudall, line 166: Marketing matters ;) but yeah OK.

README.md
```diff
@@ -195,7 +195,7 @@ client.chat(
 # => "Anna is a young woman in her mid-twenties, with wavy chestnut hair that falls to her shoulders..."
 ```
 
-Note: the API docs state that token usage is included in the streamed chat chunk objects, but this doesn't currently appear to be the case. To count tokens while streaming, try `OpenAI.rough_token_count` or [tiktoken_ruby](https://github.com/IAPark/tiktoken_ruby).
+Note: OpenAPI currently does not report token usage for streaming responses. To count tokens while streaming, try `OpenAI.rough_token_count` or [tiktoken_ruby](https://github.com/IAPark/tiktoken_ruby). Also, each call to the stream proc corresponds to a single token, so you can count the number of calls to the proc to get the completion token count.
```
alexrudall, line 198: I believe you, but is there a source for this in docs somewhere? Not heard it before.

alexrudall (commented):
Thanks a lot for this!

README.md
atesgoral: I haven't seen this anywhere in their documentation, but after months and months of working with OpenAI, I have confidently observed this to be the case. I'll ask OpenAI.

atesgoral:
Ah, I just remembered removing that HTTP error case from the http spec because I was running into issues with mocking a request. I then added a VCR at the client spec level instead.

The old test might have popped back in due to a bad rebase. I removed it again.

alexrudall (approved):

Issue #296, opened 2023-08-03
Deprecation of v1/completions endpoint


chasemcdo:
**Is your feature request related to a problem? Please describe.**
In January the v1/completions endpoint will be deprecated: https://platform.openai.com/docs/deprecations

**Describe the solution you'd like**
Enable completion with gpt-3.5+ based models

**Additional context**
With the upcoming deprecation of endpoints that this package is reliant on for completion, it is approaching the time where we should consider enabling completion with non-deprecated models.

alexrudall:
Thanks @chasemcdo - Completions endpoint removed in [v6](https://github.com/alexrudall/ruby-openai/releases/tag/v6.0.0)

PR #350, opened 2023-10-31
Remove deprecated completions endpoint

alexrudall:
- [Completions endpoint is deprecated](https://openai.com/blog/gpt-4-api-general-availability) and will be removed on 4 Jan 2024


chasemcdo:
Resolves #296

PR #351, opened 2023-11-05
Fix Maintain Unchanged Hash Parameter in Method json_post

martinjaimem:
## Description:

When calling the `chat` method, the parameters used when calling it are being altered, specifically the parameters[:stream] entry:

```ruby
parameters = {
     model: "gpt-3.5-turbo", # Required.
     messages: [{ role: "user", content: "Describe a character called Anna!"}], # Required.
     temperature: 0.7,
         print chunk.dig("choices", 0, "delta", "content")
     end
}
client.chat(parameters: parameters)

puts parameters[:stream]
# true
```

This can cause bugs as parameters should remain unchanged, we noticed it when we tried retrying the request on some OpenAi network failures, we called the method again with the same parameters and noticed streaming wouldn't work then.

**Impact**: This change does not affect the external behavior of the chat method but ensures parameter integrity across multiple calls.

This PR was made in collaboration with @vickymadrid03 and @nicastelo 👏 

### DIFF
--- a/lib/openai/http.rb
+++ b/lib/openai/http.rb
@@ -10,15 +10,7 @@ module OpenAI
 
     def json_post(path:, parameters:)
       to_json(conn.post(uri(path: path)) do |req|
-        if parameters[:stream].respond_to?(:call)
-          req.options.on_data = to_json_stream(user_proc: parameters[:stream])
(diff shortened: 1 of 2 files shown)


alexrudall (approved):
This is great, thanks!

PR #355, opened 2023-11-07
enable frozen string literals for performance and thread safety

ryanseys:
We should enable frozen string literals by default for performance and thread safety.

I got GPT-4 to explain why they are important:

Screenshot 2023-11-07 at 12 23 23 AM

Less important for new versions of Ruby, but also eliminates a bunch of `.freeze` sprinkled in the code too, which is nice. And Rubocop will automatically add them.

## All Submissions:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,4 +28,4 @@ Style/StringLiterals:
   EnforcedStyle: double_quotes
 
 Style/FrozenStringLiteralComment:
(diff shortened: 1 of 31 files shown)


alexrudall:
Good change, thanks! Will try and fix the specs. ChatGPT is wrong about Ruby 3 default though, Matz changed his mind as too disruptive

PR #356, opened 2023-11-07
Adds OpeAI::AssistantFiles

codenamev:
This adds support for the new [OpenAI Assistant Files API](https://platform.openai.com/docs/api-reference/assistants/file-object).

It is dependent on #361.  I have rebased off of that.


alexrudall:
That was fast :D

Oldharlem:
Missing threads and messages for the complete assistant flow. 
Will work on it tomorrow if not already on it.

etewiah:
Looks like there is this too which is aiming for the same goal:

https://github.com/alexrudall/ruby-openai/pull/361

Haegin:
> Looks like there is this too which is aiming for the same goal:
> 
> #361

Yeah sorry, I didn't see this PR until I'd already written most of mine. I've not added any of the file handling stuff though so it'd be good to add your changes in as you're further ahead there. Are you okay with that?

codenamev:
> > Looks like there is this too which is aiming for the same goal:
> > #361
> 
> Yeah sorry, I didn't see this PR until I'd already written most of mine. I've not added any of the file handling stuff though so it'd be good to add these changes in as you're further ahead there. Are you okay with that @codenamev?

Woah, #361  is _way_ better than this 😆  Haha, yeah for sure, I'll update to focus on assistant file updates.

codenamev:
@alexrudall I don't really know how to fix the VCR issues 🆘

lib/openai/assistant_files.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module OpenAI
+  class AssistantFiles
+    class InvalidFileFormat < OpenAI::Error; end
+
+    VALID_EXTENSIONS = %w[
```
allcentury, line 7: I think validating at the client level here is a going to be game of whack-a-mole and we should let the server enforce this instead.

lib/openai/assistant_files.rb
```diff
@@ -0,0 +1,60 @@
...
+      @client.get(path: "/assistants/#{@assistant_id}/files")
+    end
+
+    def upload(parameters: {})
+      validate(file: parameters[:file])
+
+      @client.multipart_post(
+        path: "/files",
+        parameters: parameters.merge(
+          file: File.open(parameters[:file]),
```
allcentury, line 27: I think we should require `file` be an IO object instead of an actual file - thoughts?

lib/openai/assistant_files.rb
alexrudall: Agreed. Easier to just allow anything I think?

lib/openai/assistant_files.rb
alexrudall: I think this is a good idea, can you say more about this @allcentury ? What's the reasoning?

lib/openai/assistant_files.rb
allcentury: ya, +1

lib/openai/assistant_files.rb
allcentury: As an example, I am generating a "File" from a database query.  With this current approach, I have to actually write to a file, then pass it here only to be read again.  That said, in ruby [File#open](https://ruby-doc.org/core-2.5.1/File.html#method-c-open) and other `File` commands returns an `IO` object which is what `Net::HTTP` and other adapters want when sending over HTTP.  

My thought here is that if the caller of this should be able to:

```ruby
@client.upload(file:  File.read("my_file"))

# OR

io = IO.new
output = SomeDatabaseQuery.extract
io.puts output
io.rewind

@client.upload(file: io)
```

In this case, it works for all use cases and users don't need to write temporary files to disk if they won't want to.

lib/openai/assistant_files.rb
codenamev: Removed 😄

lib/openai/assistant_files.rb
alexrudall: @allcentury thanks for the suggestion - this has now been added in #420 💨

alexrudall:
Thanks for your work on this @codenamev  - closing as OpenAI have [deprecated AssistantFiles for v2 of the beta](https://platform.openai.com/docs/assistants/migration/what-has-changed)

PR #358, opened 2023-11-07
Handle case where error response body may not be JSON

### LINKED ISSUE #357: HTTP responses without a JSON body will raise JSON parser exceptions
**Describe the bug**
With #344 there's an oversight: When the OpenAI fetch fails without a JSON error in the body, it will still result in an error being raised, but it will be a JSON parse exception.

**To Reproduce**
Tweak spec/fixtures/cassettes/gpt-3_5-turbo_streamed_chat_with_error_response.yml to return a 500 and remove the `body` (or make it invalid JSON).

**Expected behavior**
A Faraday Internal Server Error should be raised with the original response body.

**Additional context**
I broke this. I'm fixing it. PR incoming soon.

atesgoral:
Fixes #357

Gracefully handle the case where an HTTP error response may not have valid JSON in its body e.g.:
- OpenAI is having a bad day an returns a 500 Internal Server Error
- Some other network element like a proxy is returning some 4xx or 5xx.

## All Submissions:

### DIFF
--- a/spec/fixtures/cassettes/gpt-3_5-turbo_streamed_chat_with_json_error_response.yml
+++ b/spec/fixtures/cassettes/gpt-3_5-turbo_streamed_chat_with_json_error_response.yml
@@ -19,11 +19,24 @@ http_interactions:
           - Ruby
     response:
       status:
-        code: 500
-        message: Internal Server Error
+        code: 400
+        message: Bad Request
       headers:
         Date:
           - Mon, 14 Aug 2023 15:02:13 GMT
+        Content-Type:
+          - application/json
+      body:
+        encoding: UTF-8
+        string: |+
+          {
+              "error": {
+                  "message": "Test error",
+                  "type": "test_error",
+                  "param": null,
+                  "code": "test"
+              }
+          }
     recorded_at: Mon, 14 Aug 2023 15:02:13 GMT
 
 recorded_with: VCR 6.1.0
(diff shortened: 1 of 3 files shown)


spec/fixtures/cassettes/gpt-3_5-turbo_streamed_chat_with_json_error_response.yml
atesgoral, line 1: Renamed the old file by inserting a _json_ in the filename

spec/fixtures/cassettes/gpt-3_5-turbo_streamed_chat_with_error_response.yml
atesgoral, line 1: This is a copy-paste-modify of the JSON error response. I just changed the status code to 500 and removed the body.

spec/openai/client/chat_spec.rb
```diff
@@ -94,6 +94,21 @@ def call(chunk)
...
+
+          context "with an error response without a JSON body" do
+            let(:cassette) { "#{model} streamed chat with error response".downcase }
+
+            it "raises an HTTP error" do
+              VCR.use_cassette(cassette, record: :none) do
+                response
+              rescue Faraday::ServerError => e
+                expect(e.response).to include(status: 500)
+                expect(e.response[:body]).to eq("")
```
atesgoral, line 106: 🎉

codenamev (approved):
Nice 👍 Thank you for addressing this!  Working as expected so far.

fabioxgn (approved):

alexrudall:
Thanks for this!

PR #361, opened 2023-11-08
Support for Assistants, Threads, Messages and Runs

Haegin:
## All Submissions:

There's a draft PR open for Assistants (#356), and that seems to include some useful stuff for files I haven't included, but it's missing threads, messages and runs. 
This adds support for the new Assistant APIs, with messages, threads, and runs to allow users to build interactive assistants/chatbots.

~I've tried to keep to project conventions, though did add `pry-byebug` as an upgrade to `byebug` for easier debugging while I was working on the project. Happy to remove that if you'd prefer.~ `pry-byebug` doesn't support Ruby 2.6 and while it's no longer supported, I don't want to force a major version bump for this change so I've removed it.


lib/openai/client.rb
```diff
@@ -53,12 +53,34 @@ def models
...
       json_post(path: "/moderations", parameters: parameters)
     end
 
     def azure?
       @api_type&.to_sym == :azure
     end
+
+    def beta(apis)
+      dup.tap do |client|
+        client.send(:add_headers, "OpenAI-Beta": apis.map { |k, v| "#{k}=#{v}" }.join(";"))
```
codenamev, line 82: `add_headers` is a public method.  Prefer calling it explicitly.
```suggestion
        client.add_headers "OpenAI-Beta": apis.map { |k, v| "#{k}=#{v}" }.join(";")
```

lib/openai/http_headers.rb
```diff
@@ -0,0 +1,34 @@
...
+        "api-key" => @access_token
+      }
+    end
+
+    def extra_headers
+      @extra_headers ||= {}
+    end
+
+    def add_headers(headers)
+      @extra_headers = extra_headers.merge(headers.transform_keys(&:to_s))
```
codenamev, line 31: Did you intend to override the extra headers here?  Seems non-obvious that this would happen if I set them explicitly on initialization.

lib/openai/threads.rb
```diff
@@ -0,0 +1,23 @@
+module OpenAI
+  class Threads
+    def initialize(client:)
+      @client = client.beta(assistants: "v1")
```
codenamev, line 4: ❤️ this beta flag!

lib/openai/assistants.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def list
+      @client.get(path: "/assistants")
+    end
+
+    def retrieve(id:)
+      @client.get(path: "/assistants/#{id}")
+    end
+
+    def create(parameters: {})
```
cjavdev, line 15: Thanks for putting this together! Could we also add a `modify` method? 

```rb
def modify(id:, parameters: {})
  @client.json_post(path: "/assistants/#{id}", parameters: parameters)
end
```

Gemfile.lock
lorismaz, line 1: It looks like the pry-byebug gem was removed from the Gemfile, but the lock file still has the dependency listed. Maybe we should update that to keep things in sync?

alexrudall:
@Haegin thanks for your work here! If you address the above comments I'll look to release this ASAP

spec/openai/client/client_spec.rb
```diff
@@ -1,3 +1,5 @@
+require "byebug"
+
```
alexrudall, line 2: Can be removed

lib/openai/assistants.rb
Haegin: Yup, I've added that for messages, threads, assistants and runs.

lib/openai/client.rb
Haegin: Good point. I didn't actually intend to make the other header methods public so I've changed the others back to private.

lib/openai/http_headers.rb
Haegin: Yes - that seemed the easiest way to add headers without changing the existing API too much. I think it's reasonable to expect that if you do something like this the latest value should take precedence. Is this what you're worried about or is there something else I'm missing here?

```ruby
client = OpenAI::Client.new(extra_headers: { 'My-Header': 'a value' })
client.add_headers('My-Header': 'another value'))
```

Haegin (commented):
Thanks for the feedback, I think I've addressed everything.

jbeker:
I'm not sure if this should be addressed in this PR or another, but the code in file uploads that tries to validate the uploaded file as JSON prevents the ability to upload arbitrary files for use with Assistants.

codenamev:
> I'm not sure if this should be addressed in this PR or another, but the code in file uploads that tries to validate the uploaded file as JSON prevents the ability to upload arbitrary files for use with Assistants.

I've added Assistant File support in #356

lib/openai/http_headers.rb
alexrudall: Yep exactly that. `extra_headers.merge(@extra_headers || {}).merge(headers.transform_keys(&:to_s))` or similar should do it I think?

lib/openai/runs.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def cancel(id:, thread_id:)
+      @client.post(path: "/threads/#{thread_id}/runs/#{id}/cancel")
+    end
+
+    def submit_tool_outputs(thread_id:, run_id:, parameters: {})
+      @client.json_post(path: "/threads/#{thread_id}/runs/#{run_id}/submit_tool_outputs",
+                        parameters: parameters)
+    end
```
geekdreamzz, line 30: perhaps we can figure this out in follow-up since the API itself is in BETA and shouldn't introduce any breaking changes. But the API ref documents essentially "nested" a step model inside of the runs model: https://platform.openai.com/docs/api-reference/runs/getRunStep ....it looks like those APIs are not exposed here in this iteration of the runs.rb ...in order to build these Assistants end-to-end we will need to get the Step Objects incorporated. (https://platform.openai.com/docs/api-reference/runs/step-object) .... but I don't have a good solution, just an observation that OPENAI didn't isolate steps into it's own class of API's...they just appended more API functionality to their runs model. Generally the PR LGTM. maybe I can start a PR once this goes in? or If you've pretty much have this going already lmk & I can just wait for those next round of changes. great work... 

Either way...the step objects are missing in this iteration? or maybe I'm just not seeing it. IMO this PR looks great and can be merged. In follow-up we need to add the step objects

jcarrpries (approved):
Changes look good to me! Can't wait to use them

lib/openai/assistants.rb
cjavdev: Thank you!

spec/openai/client/client_spec.rb
Haegin: Thanks. Gone!

lib/openai/runs.rb
Haegin: Good point - I've added support for them as a separate class.

```ruby
client.run_steps.list(thread_id: thread_id, run_id: run_id)
client.run_steps.retrieve(thread_id: thread_id, run_id: run_id, id: step_id)
```

jcarrpries (approved):

jcarrpries:
Can we get a final approval and merge on this @alexrudall Looking to add this support into my project asap.

swistaczek:
@alexrudall I also reviewed it on my end, and I think this looks great. Let us know when you think we can access this one upstream. Thank you!

swistaczek (approved):

alexrudall:
Big thanks all - released in [v6.1](https://github.com/alexrudall/ruby-openai/releases/tag/v6.1.0)

dandailey:
Excited to use this! Any docs/info on how?

alexrudall:
@dandailey PRs welcome on adding these endpoints to the README :D https://github.com/alexrudall/ruby-openai/issues/376

alexrudall:
@dandailey basically you make an assistant like [this](https://github.com/alexrudall/ruby-openai/blob/main/spec/openai/client/assistants_spec.rb) and then run it like [this](https://github.com/alexrudall/ruby-openai/blob/main/spec/openai/client/runs_spec.rb)

PR #373, opened 2023-11-15
Add ability to pass Faraday config block to Client

alexrudall:
- Add test that error messages are included in Faraday response body
- Add ability to pass Faraday config block to client

## All Submissions:

### DIFF
--- a/spec/openai/client/chat_spec.rb
+++ b/spec/openai/client/chat_spec.rb
@@ -31,38 +31,37 @@ RSpec.describe OpenAI::Client do
           let(:messages) do
             [
               {
-                "role"=>"function",
+                "role" => "function",
                 # "name" => "function",
-                "content"=>"function"
+                "content" => "function"
               }
             ]
           end
           let(:functions) do
             [
               {
-                "name"=>"function",
-                "description"=>"function",
-                "parameters"=>
+                "name" => "function",
+                "description" => "function",
+                "parameters" =>
                   {
-                    "type"=>"object",
-                    "properties"=>{
-                      "user"=>{
-                        "type"=>"string",
-                        "description"=>"the full name of the user"
+                    "type" => "object",
+                    "properties" => {
+                      "user" => {
+                        "type" => "string",
+                        "description" => "the full name of the user"
                       }
-                    },
+                    }
                   }
-              },
+              }
             ]
           end
 
           it "raises an error containing the reason" do
             VCR.use_cassette(cassette) do
-              begin
-                response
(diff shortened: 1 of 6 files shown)


spec/openai/client/chat_spec.rb
```diff
@@ -24,6 +26,46 @@
...
+
+          it "raises an error containing the reason" do
+            VCR.use_cassette(cassette) do
+              response
+            rescue Faraday::Error => e
+              expect(e.response.dig(:body, "error",
+                                    "message")).to include("Missing parameter 'name'")
+            end
+          end
+        end
```
alexrudall, line 67: This spec is kind of unrelated but being unable to find this error is what triggered this PR. So including it

ErikDeBruijn:
This is such interesting timing. I was thinking of adding something along these lines, actually, because so far I was monkey patching this in the initializer `openai.rb`!

```
OpenAI.configure do |config|
  config.access_token = Rails.application.credentials.open_ai_key
  config.request_timeout = 300
end

module OpenAI
  module HTTP
    def connection_with_timeout(multipart: false)
      original_connection(multipart:).tap do |conn|
        conn.use TimeoutMiddleware, timeout: @request_timeout
      end
    end

    alias_method :original_connection, :conn
    alias_method :conn, :connection_with_timeout
  end
end
```

Obviously, having such a block to configure the Faraday object is an improvement.

FYI: this was because I was still observing time outs

alexrudall:
> This is such interesting timing. I was thinking of adding something along these lines, actually, because so far I was monkey patching this in the initializer `openai.rb`!
> 
> 
> 
> ```
> 
> OpenAI.configure do |config|
> 
>   config.access_token = Rails.application.credentials.open_ai_key
> 
>   config.request_timeout = 300
> 
> end
> 
> 
> 
> module OpenAI
> 
>   module HTTP
> 
>     def connection_with_timeout(multipart: false)
> 
>       original_connection(multipart:).tap do |conn|
> 
>         conn.use TimeoutMiddleware, timeout: @request_timeout
> 
>       end
> 
>     end
> 
> 
> 
>     alias_method :original_connection, :conn
> 
>     alias_method :conn, :connection_with_timeout
> 
>   end
> 
> end
> 
> ```
> 
> 
> 
> Obviously, having such a block to configure the Faraday object is an improvement.
> 
> 
> 
> FYI: this was because I was still observing time outs

Ah sweet - could you test this PR with that use case? Would be much appreciated 🙏

obie (approved):
Please do merge

PR #388, opened 2023-11-26
Teach runs list endpoints to use parameters

shalecraig:
## Change

Teach the `client.runs.list` endpoint about a `parameters:` kwarg. That kwarg is passed to the underlying GET request.

## Motivation

I would like to check the latest run for a thread, and it seems like the best way to do this is to list the runs for a thread limited to 1.

## All Submissions:

### DIFF
--- a/lib/openai/runs.rb
+++ b/lib/openai/runs.rb
@@ -4,8 +4,8 @@ module OpenAI
       @client = client.beta(assistants: "v1")
     end
 
-    def list(thread_id:)
-      @client.get(path: "/threads/#{thread_id}/runs")
+    def list(thread_id:, parameters: {})
+      @client.get(path: "/threads/#{thread_id}/runs", parameters: parameters)
     end
 
     def retrieve(thread_id:, id:)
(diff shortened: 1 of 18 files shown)


lib/openai/runs.rb
```diff
@@ -4,8 +4,8 @@ def initialize(client:)
       @client = client.beta(assistants: "v1")
     end
 
-    def list(thread_id:)
-      @client.get(path: "/threads/#{thread_id}/runs")
+    def list(thread_id:, parameters: {})
```
shalecraig, line 7: Ideally I'd make matching changes to `runs_spec.rb` -- are there ways that you generate the cassette? (I'm looking to avoid doing this manually)

alexrudall:
Thanks for this, could you add this to the Messages#list endpoint also?

coezbek:
@alexrudall If you pull @shalecraig PR, I can send you another PR with the messages also updated.

I send @shalecraig the PR already, but am not sure if he sees it: https://github.com/coezbek/ruby-openai/tree/patch-1

coezbek:
This is also relevant for RunSteps#list endpoint, even though I doubt there should be more than 20 items in this end-point.

alexrudall:
Thanks @shalecraig & @coezbek!

PR #391, opened 2023-11-27
Files for assistants can now be PDFs,... as we only validate .jsonl files

stefan-kp:
Currently all files are validated on upload so it is impossible to use PDFs for assistants
This small fix now allows other types


alexrudall:
Thanks @stefan-kp!

Issue #392, opened 2023-12-01
Only Require Faraday Multipart if Faraday Version >= 2


ajGingrich:
**Describe the bug**

After upgrading from 3.7.0 to the latest version, I see warnings related to already initialized constants from Faraday Multipart. This is because I'm still using Faraday 1. 

**To Reproduce**

1. Use Faraday 1.1
2. Use latest ruby-openai
3. See warnings below

```
/usr/local/bundle/gems/faraday-multipart-1.0.4/lib/faraday/multipart.rb:15: warning: already initialized constant Faraday::FilePart
/usr/local/bundle/gems/faraday-1.1.0/lib/faraday/file_part.rb:53: warning: previous definition of FilePart was here
```

**Expected behavior**

Only require faraday multipart if the Faraday version is >= 2 so the constants are not required twice.

**Screenshots**

![image](https://github.com/alexrudall/ruby-openai/assets/22775484/b284b083-e9ca-4b8c-a316-4fed80600ab4)

**Additional context**

It should be a one line fix here https://github.com/alexrudall/ruby-openai/blob/80da19351b6920309a52b95e18897e23fdf72166/lib/openai.rb#L2

```ruby
require 'faraday/multipart' if Gem::Version.new(Faraday::VERSION) >= Gem::Version.new('2.0')
```

I can make the change if desired.

alexrudall:
Thanks @ajGingrich, fixed in [ruby-openai v8](https://rubygems.org/gems/ruby-openai)

PR #393, opened 2023-12-01
Conditionally require faraday/multipart

ajGingrich:
Only Require Faraday Multipart if Faraday Version is greater than 2 because Faraday V1 includes the gem already.

https://github.com/alexrudall/ruby-openai/issues/392

## All Submissions:


alexrudall:
@ajGingrich I would definitely prefer that as it could affect tens of thousands of people if we get it wrong :) I always try to test PRs as much as possible even small ones. Thank you!

alexrudall:
eep, you're right...

I think we need to require faraday json response and request middleware, but can't find the right library 🤔 

Thanks for checking, that's a really good spot

PR #394, opened 2023-12-02
Call user proc when stream is done

solodolo:
## Description
This PR sends an empty Hash to the provided user proc when the final termination message is received from the OpenAI API. 

## Why?

My use-case for this is cleanup at the end of the stream, specifically flushing a buffer. For example:

```ruby
def stream_proc
  buffer = ''
  buffer_size = 10
  proc do |chunk|
    chunk_content = chunk.dig("choices", 0, "delta", "content")
    buffer += chunk_content if chunk_content
    flush = buffer.length >= buffer_size

    # Currently no way to flush the buffer on stream termination so some data might still be in the buffer.
    # With this PR, something like the following is possible:
    # flush = buffer.length >= buffer_size || chunk == {}
    if flush
      # Do the thing with buffer
      buffer = ''
    end
  end
end
```

I am interested to hear folks' thoughts on this problem and my particular solution. I'm a little ambivalent about the latter because it is a breaking change for those that expect the current behavior, but I tried a couple different ways to avoid that and it got messy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ puts response.dig("choices", 0, "message", "content")
 
 [Quick guide to streaming Chat with Rails 7 and Hotwire](https://gist.github.com/alexrudall/cb5ee1e109353ef358adb4e66631799d)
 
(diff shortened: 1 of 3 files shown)


alexrudall:
Hmm. I think if the buffer is created outside the stream_proc, you can flush it from there, after the call to OpenAI finishes?

alexrudall:
Nice. If it wasn't a breaking change I'd be open to this, it is a shame that we lose the [DONE] part. Maybe something for the next major version

PR #395, opened 2023-12-04
Updated docs to cover base uris that contain api_version already 

kaiwren:
We're using Portkey. Portkey's base uri is `https://api.portkey.ai/v1/proxy/`. 

To use this base uri correctly required setting `config.api_version = ''` to avoid generating a bad URI like `https://api.portkey.ai/v1/proxy/v1/chat/completions`. I've updated the README to reflect this.


alexrudall:
Thanks, @kaiwren. What about fixing this by adding some code to check for the version in the path, and skipping it if it's there?

alexrudall:
Sorry, I rebased this and delete your commit @kaiwren - added code to handle it instead of README change in #460

Issue #405, opened 2023-12-12
The README needs an index


alexrudall:
**Is your feature request related to a problem? Please describe.**
The README is a bit overwhelming and it's hard to see what in it without reading the whole thing

**Describe the solution you'd like**
- A simple index of the main sections near the top
- Should be clickable to take you the right part of the README

PR #406, opened 2023-12-15
Add log_errors option to enable/disable logging of errors.

lalunamel:
## All Submissions:

Hello - first contribution here! 👋 

**This PR adds the `log_errors` option, which enables/disables the `MiddlewareErrors` middleware, controlling whether or not errors are logged when they are encountered**

The PR includes:
- changes to the README
- changes to the implementation
- changes to the tests

**Motivation:**
I'd like to use `ruby-openai` because it looks like a wonderful, well-maintained project. Unfortunately, the library logs errors when they are encountered. The information logged includes the response body sent over from OpenAI, which might conceivably include part of the prompt or response. In order to avoid leaking that information to the system orchestrating the request, I'd like to disable logging of errors.

**Testing:**
I've run all the rspec tests locally and tested the `log_errors` functionality against my own local project to confirm it works and does not impact behavior anywhere else.

Please let me know if there are any changes you'd like to see in the PR before it can be accepted!


README.md
```diff
@@ -117,7 +118,19 @@ client = OpenAI::Client.new(access_token: "access_token_goes_here")
 client.add_headers("X-Proxy-TTL" => "43200")
 ```
 
-#### Verbose Logging
```
lalunamel, line 120: I also have my eye on https://github.com/alexrudall/ruby-openai/issues/405 😄

lib/openai/client.rb
```diff
@@ -17,7 +18,10 @@ def initialize(config = {}, &faraday_middleware)
       CONFIG_KEYS.each do |key|
         # Set instance variables like api_type & access_token. Fall back to global config
         # if not present.
-        instance_variable_set("@#{key}", config[key] || OpenAI.configuration.send(key))
```
lalunamel, line 20: Turns out the statement `config[key] || OpenAI.configuration.send(key)` will always return the right hand side if the left hand side is `false`, which I don't think is the intention here.

alexrudall:
Thanks for this! I'd like to make it a bit more powerful - what about instead adding a `error_response:` param, making that default to the current logic (ie everything in the rescue in MiddlewareErrors), and then people can pass false to that (your usecase), OR their own custom logic as a proc with how to handle Faraday::Errors. So they could make their own logic for logging errors easily.

lib/openai/client.rb
alexrudall: eep good catch.

lalunamel:
Thanks for your quick reply!

I agree with your sentiment about not logging errors by default. Didn't want to be too disruptive in my first PR here, but if you approve, then 👍 . Added a commit to change the behavior.

If you are 👍 on this PR, feel free to approve and merge.

lalunamel:
Hi @alexrudall - any thoughts on this?

lalunamel:
Ping @alexrudall - let me know if there's anything else to do here!

alexrudall:
Thanks so much for this @lalunamel and apologies for the delay in response. I consider this a breaking change and will merge it soon in next major release.

lalunamel:
Woo!

PR #424, opened 2024-01-22
Remove inexistent threads#list API

ignacio-chiazzo:
The [Threads API](https://platform.openai.com/docs/api-reference/threads/createThread) doesn't expose a method to get the list of threads. It isn't listed on the official documentation:

Screenshot 2024-01-21 at 10 10 17 PM

Trying that endpoint returns the following error:
```
"OpenAI HTTP Error (spotted in ruby-openai 6.3.1): {"error"=>{"message"=>"Your request to GET /v1/threads must be made with a session key (that is, it can only be made from the browser). You made it with the following key type: secret.", "type"=>"invalid_request_error", "param"=>nil, "code"=>"missing_scope"}}"
```

## All Submissions:

### DIFF
--- a/lib/openai/threads.rb
+++ b/lib/openai/threads.rb
@@ -4,10 +4,6 @@ module OpenAI
       @client = client.beta(assistants: "v1")
     end
 
-    def list
-      @client.get(path: "/threads")
-    end
-
     def retrieve(id:)
       @client.get(path: "/threads/#{id}")
     end


alexrudall:
Thanks, @ignacio-chiazzo - looks like it was undocumented and [removed in November](https://community.openai.com/t/python-package-missing-option-to-list-all-threads/487892/12). I'm not call this a breaking change as it was not documented and was gone within a week of this method being added.

PR #426, opened 2024-01-24
Update README.md - mention "language" attribute explicitely in #transcribe

nfedyashev:
I believe that some significant portion of `#transcribe` function might be using it in languages other than English.
For them it might be handy to clearly see that "language" attribute is (implicitly) supported so that they don't need to check `Create transcription` documentation at https://platform.openai.com/docs/api-reference/audio/createTranscription?lang=node.

[X] I checked locally that it works with explicit "en" value and a few other languages

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -778,6 +778,7 @@ response = client.audio.transcribe(
     parameters: {
         model: "whisper-1",
         file: File.open("path_to_file", "rb"),
+        language: "en"
     })
 puts response["text"]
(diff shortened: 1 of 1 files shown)


alexrudall:
Thanks, @nfedyashev

alexrudall:
Released in [v6.4](https://github.com/alexrudall/ruby-openai/releases/tag/v6.4.0)

PR #436, opened 2024-03-13
Allow usage of offline models with Ollama

ThomasSevestre:
Some offline models are OpenAI compatible (ollama for instance : https://ollama.com/blog/openai-compatibility)

Offline models do not require an access token.

This PRs makes access token optional in order to be able to use self hosted models

## All Submissions:


khalilgharbaoui:
@alexrudall can this be merged?

alexrudall:
I'm open to this but would like some instructions on how to actually do this in the README. Like, a link to how to run Ollama locally and then exactly what config is needed. Also, that error is I think helpful for people who just want to use OpenAI API - might be better to pass an option like `api_type: :ollama` to skip the error

ThomasSevestre:
I've added a section in the readme to explain how to use this gem with ollama

I believe we should not add a new option in the configuration, keeping the same API allow gem using this gem to work without any change.

Here is the behavior of this branch when we try to connect to OpenAPI without an acess_token :

```ruby
client = OpenAI::Client.new
response = client.chat(
  parameters: {
    model: "mistral:latest", # Required.
    messages: [{ role: "user", content: "Hello!"}], # Required.
    temperature: 0.7,
})
# => OpenAI HTTP Error (spotted in ruby-openai 6.3.1): {"error"=>{"message"=>"You didn't provide an API key. You need to provide your API key in an Authorization header using Bearer auth (i.e. Authorization: Bearer YOUR_KEY), or as the password field (with blank username) if you're accessing the API from your browser and are prompted for a username and password. You can obtain an API key from https://platform.openai.com/account/api-keys.", "type"=>"invalid_request_error", "param"=>nil, "code"=>nil}}
```

alexrudall:
Thanks @ThomasSevestre - after testing I went with your original idea and removed the access_token check - appreciate your work on this!

Issue #452, opened 2024-04-12
Documentation does not include tool choice option


Jbrito6492:
**Describe the bug**

The README example of using functions does not mention the tool choice parameter, which intermittently caused issues during implementation. Sometimes, the API would use our function, resulting in a `finish_reason` being `tool_calls`. Other times, the API would not use our function, with a `finish_reason` being `stop`.

**To Reproduce**
Follow the functions example in the README.

**Expected behavior**
We expected the function call to be used consistently with a `finish_reason` of `tool_calls`.

**Screenshots**
N/A

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

alexrudall:
Released in [v7](https://github.com/alexrudall/ruby-openai/releases/tag/v7.0.0)

PR #453, opened 2024-04-12
Update function calling section of README

Jbrito6492:
This change addresses the [issue](https://github.com/alexrudall/ruby-openai/issues/452).  `tool_calls` is an array, so we must access the first index when retrieving values.  Furthermore, the `tool_choice` key was left out of the documentation, causing a variance in behavior.

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,9 +338,10 @@ You can stream it as well!
 
 ### Functions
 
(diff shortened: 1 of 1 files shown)


alexrudall:
Thanks @Jbrito6492

PR #473, opened 2024-05-07
If being used for Groq, skip OpenAI-Organization header

drnic:
I've looked at using groq via langchain -> ruby-openai and was seeing a warning about me not setting `OpenAI-Organization`. If I run `.compact` on the headers, the warning seems to go away.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@
 /test/tmp/
 /test/version_tmp/
 /tmp/
-/.bundle/
 
 # Used by dotenv library to load environment variables.
 .env
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@
 /.yardoc
 /_yardoc/
 /doc/
-/TAGS
 
 # Used by dotenv library to load environment variables.
 .env
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@
 /.yardoc
 /_yardoc/
 /doc/
-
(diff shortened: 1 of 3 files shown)


drnic:
^ is this CI failure something I can fix?

alexrudall:
@drnic no, I think I broke something. Will try and fix. Thanks for this PR!

.gitignore
```diff
@@ -14,7 +14,7 @@
 /.yardoc
 /_yardoc/
 /doc/
-
+/TAGS
```
alexrudall, line 17: What's TAGS?

alexrudall:
> ^ is this CI failure something I can fix?

I may have fixed this, if you push another change let's see

.gitignore
drnic: Ugh, it’s one of those metadata files that are spat out by a vscode plugin I have. 

Perhaps there’s a way I can have it in a global ignore on my machine?

alexrudall: Yeah you can https://stackoverflow.com/questions/45333484/how-to-use-global-gitignore-with-visual-studio-code

drnic: Cool. You ok to yank this change from the PR? Or I can go find my laptop and do it

alexrudall:
Thanks again @drnic - released in [v7.1](https://github.com/alexrudall/ruby-openai/releases/tag/v7.1.0)

Issue #450, opened 2024-04-04
client.files.content(id: "file-123") failing for image_files because of JSON parse


Lasvad:
**Describe the bug**
Reading files such as pngs that are generated by code interpreter seem to fail as the response is expected to be JSON

Doc reference this to read file content: `client.files.content(id: "file-123")`

example

Code interpreter returns... 
```
"content"=>
  [{"type"=>"image_file", "image_file"=>{"file_id"=>"file-stub"}},
   {"type"=>"text",
    "text"=>
     {"value"=>
```

using the cmd
```
client.files.content(id: "file-stub")
```

returns
```
(JSON::ParserError)s/ruby-3.2.2/lib/ruby/3.2.0/json/common.rb:216:in `parse': unexpected token at '\x89PNG
```

I've tried to use File.write, TempFile, StringIO but the JSON parse error continues. 

**To Reproduce**
1. Use Assistant to create a png with code interpreter 
2. Try to read the file_id that is provided in the response
3. Error shows when using `client.files.content(id: "file-123")`

**Expected behavior**
Need a way to parse image_files files from OpenAI

**Screenshots**
<img width="1272" alt="image" src="https://github.com/alexrudall/ruby-openai/assets/9116207/20e85019-cc08-41aa-aa94-aac575261e2c">

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser N/A

**Additional context**
ruby 3.2.2
Rails 7.0.4.3

DanL12186:
[Duplicate](https://github.com/alexrudall/ruby-openai/issues/421) of this PR

alexrudall:
Closing for #421

Issue #421, opened 2024-01-19
Content retrieval (and anything calling .get) only works for JSON


DanL12186:
**Describe the bug**
Attempting to retrieve the content of files via the API doesn't work unless the files are in `JSON` format; you'll get a `JSON::ParserError` error, because `get()` in `http.rb` calls `jsonparsel()` which calls `JSON.parse(response)` for every request. 

For example, when trying to retrieve an image so we can attach it via ActiveStorage in order to easily display it to the user:

```ruby
>> client.files.content(id: "file-xyz123")
JSON::ParserError: 451: unexpected token at '?PNG
```

Simply returning the response if it's not JSON would resolve this.

**To Reproduce**
1. Have the API generate a file for you
2. Call `client.files.content(id: file_id)`
3. See error

**Expected behavior**
It should retrieve non-JSON content like generated PNGs successfully

**Desktop (please complete the following information):**
 - OS: MacOS X 10.15.7
 - Browser N/A
 - Version N/A

leoplct:
Same issue.
The API supports the parameter purpose="assistants" and accepts several formats.

UPLOAD FILE: https://platform.openai.com/docs/assistants/tools/passing-files-to-code-interpreter
Accepted formats: https://platform.openai.com/docs/assistants/tools/supported-files

Here the library code: https://github.com/alexrudall/ruby-openai/blob/main/lib/openai/files.rb

thisismydesign:
Having the same issue trying to send files. No matter the type of file I'm trying to send it tries to JSON parse it.

```
        file = entity.attachment.download
        temp_file = Tempfile.new(['email', '.html'])
        temp_file.write(file)
        temp_file.rewind

client.files.upload(parameters: { file: temp_file, purpose: "assistants" })
```

`JSON::ParserError: unexpected token`

DanL12186:
@thisismydesign Figure you've worked this out, but in case you're having issues, you can easily work around it just by using a built-in or gem-based HTTP library to make a call to the files endpoint

```ruby
RestClient.get(
  "https://api.openai.com/v1/files/#{file_id}/content",
  {Authorization: "Bearer #{ACCESS_TOKEN}"}
)
  ```

alexrudall:
I believe #566  resolves this issue - if you are able to could you give it a test @DanL12186 , @thisismydesign , @leoplct  ? Thank you!

alexrudall:
[...]

PR #474, opened 2024-05-08
Fix files content endpoint

willywg:
This PR addresses issues #421 and #450.

It fixes the `client.files.content` API endpoint. The current implementation requires parsing the response body to JSON. However, some responses (like files generated by the Assistant API) need to be handled as pure response body content, such as a PNG file.

## All Submissions:


mepatterson:
+1 since the Batches examples in the docs fail when retrieving the final content results for this same reason.

alexrudall:
Thanks @willywg @mepatterson -- in my understanding Faraday should only try to handle as json if the response header is `Content-Type: application/json` or similar -- wonder if we could do something similar so that users don't need to do anything different? In your examples OpenAI sends `application/octet-stream` so I think this could work

mepatterson:
Makes sense to me

gregszero:
i'm now facing the same issue #421 ,  this could do the trick following @alexrudall comment 

```
def get(path:, parameters: nil)
  response = conn.get(uri(path: path), parameters) do |req|
    req.headers = headers
  end

  content_type = response.headers['content-type']
  content_type == 'application/octet-stream' ? response.body : parse_jsonl(response.body)
end

```
should i open a new PR?

alexrudall:
I believe https://github.com/alexrudall/ruby-openai/pull/566 resolves this issue - if you are able to could you give it a test @mepatterson , @willywg  ? Thank you!

alexrudall:
Thanks @willywg  - fixed in [ruby-openai v8](https://rubygems.org/gems/ruby-openai)

PR #476, opened 2024-05-09
Update README audio section to show full parameter support

krschacht:
I had to poke around the code to make sure that the additional parameters for the audio endpoint were supported. They are since ruby just passes them straight on through. I figured it's worth updating the README to reflect it just like the other sections which also mention the optional parameters. This should save someone a little time in the future.


alexrudall:
Love this sort of PR, tweaks like this probably save hundreds of man-hours on aggregate. if one person has a problem likely many others will too. Thanks @krschacht

Issue #479, opened 2024-05-14
OpenAI error payloads in the body may not get parsed fully during chat streaming


atesgoral:
**Describe the bug**

With streaming chat from OpenAI, if OpenAI returns an HTTP error (e.g. 400), the error handler optimistically parses the latest chunk it received as the entire body. However, networks conditions or an intermediary like an OpenAI proxy may cause the response to be fragmented into different chunks.

The symptom is an error being raised with just the first chunk of the body. While this is not a very severe problem, since it's happens on a failure path (that should ideally be rare), not collecting and parsing the full error would make it hard for application developers to figure out exactly what went wrong.

**To Reproduce**

Hard to reproduce in vivo, unless you're running a special HTTP/TCP proxy that chunks the original TCP packets for you.

_(I'll follow up with a minimal reproducible unit test.)_

**Expected behavior**

The full response body is collected (chunks concatenated) and parsed as JSON.

ajsharp:
Fixed in #635

PR #480, opened 2024-05-14
Error body parsing

atesgoral:
Add minimal reproducible test case for #479

I'll try to follow up with a fix.

## All Submissions:


alexrudall:
Thanks very much for this, @atesgoral - did you have an idea for the fix here?

atesgoral:
@alexrudall I started looking at a fix a while back, but it didn't turn out trivial due to all the streaming / passing around of blocks. I don't remember the details. But this is something I think is still worth patching!

PR #485, opened 2024-05-20
Improve README function calling for chat

krschacht:
Expanded the README to reflect a realistic use of function calling in the chat completion API


alexrudall:
Thank you @krschacht!

PR #491, opened 2024-05-28
Add open_timeout option to avoid waiting for 120 seconds under poor internet conditions

allen12921:
## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,12 +133,13 @@ client = OpenAI::Client.new(access_token: "access_token_goes_here")
 
 #### Custom timeout or base URI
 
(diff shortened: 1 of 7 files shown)


alexrudall (commented):
Thanks for this, @allen12921. Since we can [add any Faraday middleware with a block](https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#faraday-middleware), why make this a default option? I'm open to it if you think it would improve the experience for especially junior developers, but just want to understand the logic.

BTW, adding a default different to the Faraday default (which seems to be nil/infinite) would also require a major (breaking) release, as it would change the default behaviour of ruby-openai.

PR #492, opened 2024-05-28
Add parameters support to batches.list and update README to include /v1/embeddings Batches support

### LINKED ISSUE #490: batches.list is missing after/limit query parameters support + README inaccuracy regarding batches
https://github.com/alexrudall/ruby-openai/blob/8eb793e5482feef1adf97854d615d151f226a6d6/lib/openai/batches.rb#L1-L23

Is missing support for the `after` and `limit` parameters described here:

> **after**
> string
> Optional
> A cursor for use in pagination. after is an object ID that defines your place in the list. For instance, if you make a list request and receive 100 objects, ending with obj_foo, your subsequent call can include after=obj_foo in order to fetch the next page of the list.
> 

> **limit**
> integer
> Optional
> Defaults to 20
> A limit on the number of objects to be returned. Limit can range between 1 and 100, and the default is 20.

[API Reference](https://platform.openai.com/docs/api-reference/batch/list)

---

In addition, the README incorrectly states that Batches are only available for completions, while embeddings are also supported.

**README:**

https://github.com/alexrudall/ruby-openai/blob/8eb793e5482feef1adf97854d615d151f226a6d6/README.md?plain=1#L501

**OpenAI Docs:**

> Batches start with a .jsonl file where each line contains the details of an individual request to the API. For now, the available endpoints are /v1/chat/completions (Chat Completions API) and /v1/embeddings (Embeddings API).

Source: https://platform.openai.com/docs/guides/batch/1-preparing-your-batch-file

marckohlbrugge:
Closes #490

I added a new test that creates a batch first, so we can test the `after` parameter. Without an existing batch, it would return an error.

**Edit:** Looks like there's some Rubocop linting issues. I wasn't able to get Rubocop running so perhaps @alexrudall you can take care of these as they seem auto-fixable.

**Edit 2:** Okay, I fixed the Rubocop linting issues. Only thing left to make the test pass is to record a new VCR cassette. I'm not comfortable getting my company details in there, so can you do it @alexrudall? I already checked locally, and the test pass with a valid API key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,11 +497,9 @@ puts response.dig("data", 0, "embedding")
 ```
 
 ### Batches
-
-The Batches endpoint allows you to create and manage large batches of API requests to run asynchronously. Currently, only the `/v1/chat/completions` endpoint is supported for batches.
+The Batches endpoint allows you to create and manage large batches of API requests to run asynchronously. Currently, the supported endpoints for batches are `/v1/chat/completions` (Chat Completions API) and `/v1/embeddings` (Embeddings API).
 
 To use the Batches endpoint, you need to first upload a JSONL file containing the batch requests using the Files endpoint. The file must be uploaded with the purpose set to `batch`. Each line in the JSONL file represents a single request and should have the following format:
-
 ```json
 {
   "custom_id": "request-1",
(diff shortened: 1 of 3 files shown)


alexrudall:
Thanks @marckohlbrugge !

alexrudall:
Released in [v7.1](https://github.com/alexrudall/ruby-openai/releases/tag/v7.1.0)

PR #502, opened 2024-07-17
(feat): add Faraday proxy support

danilrwx:
## Summary
Adding the ability to use a proxy when proxy from ENV is not suitable.

```ruby
proxy = {
  :uri      => 'http://proxy.example.com',
  :user     => 'foo',
  :password => 'bar'
}

OpenAI::Client.new(access_token: "access_token_goes_here", proxy: proxy)
```

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,26 @@ client = OpenAI::Client.new(access_token: "access_token_goes_here")
 client.add_headers("X-Proxy-TTL" => "43200")
 ```
 
+#### Proxy support
+
+You can use proxy with Faraday
+
+```ruby
+proxy = {
+  :uri      => 'http://proxy.example.com',
+  :user     => 'foo',
+  :password => 'bar'
+}
+
+OpenAI::Client.new(access_token: "access_token_goes_here", proxy: proxy)
+```
+
+Or use with string
+
+```ruby
+OpenAI::Client.new(access_token: "access_token_goes_here", proxy: "foo:bar@http://proxy.example.com")
+```
+
 #### Logging
 
 ##### Errors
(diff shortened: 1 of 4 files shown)


cpb:
I checked out this branch (rebasing on `alexrudall:ruby-openai main`) and it worked for me!

I noticed this branch is missing test coverage. I found a couple libraries with test coverage for the proxy configuration:
- [`sentry-ruby` tests their `URI` object support by checking the state](https://github.com/getsentry/sentry-ruby/blob/6181b7bab798cdf3b846d853de5d4b00b706dfee/sentry-ruby/spec/sentry/transport/http_transport_spec.rb#L120-L131)
- [`oktokit` also uses state based assertions](https://github.com/octokit/octokit.rb/blob/301bb57668bc1632e0cc93cbb5212a35668f4a76/spec/octokit/client_spec.rb#L439-L445)

@alexrudall would you accept this PR with or without test coverage? Or would you prefer `oktokit` style test coverage?

alexrudall:
Apologies for the delay in review. 
- @danilrwx Thanks very muchfor this PR and for adding to the README - a bit more about when/why you might want to do this could be really helpful.
- @cpb - yes, a simple 'happy path' test would be great, can probably be added to 1 of the specs in `spec/openai/client/client_spec.rb`

cpb:
>     * In which case, this could be just added to the existing README example to help users do this.

Thank for the feedback @alexrudall, see:
- https://github.com/alexrudall/ruby-openai/pull/539

danilrwx:
@alexrudall awesome, then I'll close the PR.

Issue #513, opened 2024-09-04
OpenAI::Client#inspect does not filter out access_token or organization_id


samwaree:
**Describe the bug**
When calling `OpenAI::Client#inspect`, I would expect the `access_token` (and `organization_id`) to be filtered out, but they are currently not. 

My current workaround is to override the `inspect` method to filter out sensitive info. But if you don't do this, this might cause some issues for developers if the output gets logged anywhere.

**To Reproduce**
1. Create an `OpenAI::Client` with an `access_token` and/or `organization_id`
2. Call `.inspect` on the client
3. See that the `access_token` and `organization_id` get printed out

**Expected behavior**
The output would either not contain the `access_token`/`organization_id` params, or they would be obfuscated, like this:

```ruby
#<OpenAI::Client:0xXXXXX @api_type=nil, @api_version=\"v1\", @access_token=***FILTERED***, @log_errors=true, @organization_id=***FILTERED***, @uri_base=\"https://api.openai.com/\", @request_timeout=120, @extra_headers={}, @faraday_middleware=nil>
```

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS

**Additional context**

PR #517, opened 2024-09-13
Hide sensitive info when inspecting client

samwaree:
Updates the `.inspect` method for `OpenAI::Client` to obfuscate sensitive values by replace them with `[REDACTED]`. This should prevent accidental logging of API keys, etc.

Resolves alexrudall/ruby-openai#513

## All Submissions:

### DIFF
--- a/lib/openai/client.rb
+++ b/lib/openai/client.rb
@@ -2,6 +2,7 @@ module OpenAI
   class Client
     include OpenAI::HTTP
 
+    SENSITIVE_ATTRIBUTES = %i[@access_token @organization_id @extra_headers].freeze
     CONFIG_KEYS = %i[
       api_type
       api_version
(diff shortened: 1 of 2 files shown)


lib/openai/client.rb
```diff
@@ -107,5 +107,17 @@ def beta(apis)
         client.add_headers("OpenAI-Beta": apis.map { |k, v| "#{k}=#{v}" }.join(";"))
       end
     end
+
+    def inspect
+      sensitive_attributes = %i[@access_token @organization_id @extra_headers]
```
samwaree, line 112: I included `extra_headers`, since my thinking was this may contain other API keys and it is better to hide too much than potentially show something sensitive. Let me know if others disagree

alexrudall:
Thanks @samwaree - would you say this is a breaking change? I tend towards not, surely nobody is relying on `inspect`

lib/openai/client.rb
alexrudall, line 112: I think it makes sense. Could put these in a global `SENSITIVE_ATTRIBUTES=`

alexrudall:
Thanks @samwaree !

PR #523, opened 2024-10-05
Add query params support

agamble:


alexrudall:
@agamble - thanks for this, I can't find the documentation for this on OpenAI? Can you add a link + description of what this does?

jorgemaciaspina:
I found some documentation in the
[create run section](https://platform.openai.com/docs/api-reference/runs/createRun)
That page has a link [file search tool documentation](https://platform.openai.com/docs/assistants/tools/file-search/customizing-file-search-settings) where you can see this example:
`curl -g https://api.openai.com/v1/threads/thread_abc123/runs/run_abc123/steps/step_abc123?include[]=step_details.tool_calls[*].file_search.results[*].content \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H "Content-Type: application/json" \
  -H "OpenAI-Beta: assistants=v2"`
Hope this helps.

agamble:
> @agamble - thanks for this, I can't find the documentation for this on OpenAI? Can you add a link + description of what this does?

@jorgemaciaspina has the correct docs and the exact use case we were trying to achieve. I haven't exhaustively gone through all the examples so there may be other POST actions that allow query params too, but "create run" was the one we were trying to use.

Thanks for your hard work on this useful library.

alexrudall:
Thanks @agamble & @jorgemaciaspina - branched off and added specs here #527 - will release shortly as v7.3.0

alexrudall:
Released in [v7.3.0](https://github.com/alexrudall/ruby-openai/releases/tag/v7.3.0), added an example to the README [here](https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#exploring-chunks-used-in-file-search) also. Thanks both!

PR #533, opened 2024-10-15
Add Assistant Vision example to README

alexrudall:
## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1072,6 +1072,71 @@ run_id = response['id']
 thread_id = response['thread_id']
 ```
 
+#### Vision in a thread
+
+You can include images in a thread and they will be described & read by the LLM. In this example I'm using [this file](https://upload.wikimedia.org/wikipedia/commons/7/70/Example.png):
+
+```
+require "openai"
+
+# Make a client
+client = OpenAI::Client.new(
+  access_token: "access_token_goes_here",
+  log_errors: true # Don't log errors in production.
+)
+
+# Upload image as a file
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1072,6 +1072,71 @@ run_id = response['id']
...
+# Upload image as a file
+file_id = client.files.upload(
+  parameters: {
+    file: "path/to/example.png",
+    purpose: "assistants"
+  }
+)["id"]
+
+# Create assistant
+assistant_id = client.assistants.create(
```
mtantawy, line 1097: First of all, thank you for documenting this and providing an example, that's really helpful!

If I could only make a change here that would be somehow communicating the fact that creating an assistant should be a one-time action, a user won't need to recreate the assistant for every attempt to describe or read an image, that's what threads are for

alexrudall: To be honest I treat them as pretty throwaway as well :D 

What about `# Create assistant (or grab an existing one`

mtantawy: It's true, they come for free, I like to keep them unique for organization purposes anyway

> To be honest I treat them as pretty throwaway as well :D
> 
> What about `# Create assistant (or grab an existing one)`

That sounds good, there is always room for clarifying more but that's probably out of scope of your example here

alexrudall: Alrighty, I tweaked it a bit more [here](https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#vision-in-a-thread)

Thanks for the suggestion!

PR #535, opened 2024-10-15
Quick readme tweaks

panozzaj:
Hi Alex, I saw your tweet this morning about using the OpenAI assistants API. Very exciting!

When I looked at the tweet, I noticed that it lacked syntax highlighting.

This PR has 2 commits. The first adds languages to code blocks that were missing them (except one related to an ENV file). I think the language commit is very low risk and would help with doc readability.

While I was in there, I also went through and made a second commit to try to reformat the code examples to reduce leading whitespace and consistently format. I think by doing this:

 - more of each example will be visible to folks who are viewing the README in their browser since leading whitespace is trimmed
 - the README will be easier to skim through due to the consistency of formatting
 - it will be easier to copy-paste, since most Ruby projects use 2 spaces for indentation in my experience
 
Open to not taking the last commit since it is a bit more opinionated. 

Thanks for your consideration of this PR.  📖 

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,7 +724,7 @@ response.dig("choices", 0, "message", "content")
 
 You can also capture the events for a job:
 
-```
+```ruby
 client.finetunes.list_events(id: fine_tune_id)
 ```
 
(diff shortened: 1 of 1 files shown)


alexrudall:
Thanks very much @panozzaj, this looks MUCH better!!

PR #545, opened 2024-11-04
[Feature] Realtime API from OpenAI working

### LINKED ISSUE #524: Realtime
OpenAI [added a realtime API](https://platform.openai.com/docs/guides/realtime/overview)! 

How do we add this to ruby-openai?

- As simple and easy to use as possible
- Adding as few dependencies as possible
- As reliable as possible

franpb14:
### WIP
This is a very basic PR that we can use to iterate. Right now I'm using it with ActionCable in a rails app like this:
```
class OpenAiChannel < ApplicationCable::Channel
  def subscribed
    stream_from "open_ai_channel"
    @client = OpenAI::Client.new(access_token: ENV['OPENAI_API_KEY'])
    @client.real_time.on_message do |event|
      ActionCable.server.broadcast 'open_ai_channel', { message: event.data.force_encoding('UTF-8') }
    end
    @client.real_time.connect
  end

  def send_message(data)
    @client.real_time.send_event(data['event'])
  end
end
```
In the example `data['event']` can be something like this:
```
{
  type: "response.create",
  response: {
    modalities: ["text", "audio"],
    instructions: "Please assist the user.",
  }
}
```
Maybe we could add more functions in order to facilitate [event management](https://platform.openai.com/docs/api-reference/realtime-client-events). Something like I said in [this comment](https://github.com/alexrudall/ruby-openai/issues/524#issuecomment-2452345122).

### Dependencies
- `faye-websocket`
- `eventmachine`

## All Submissions:

Closes #524

### DIFF
--- a/lib/openai/real_time.rb
+++ b/lib/openai/real_time.rb
@@ -6,9 +6,14 @@ module OpenAI
     def initialize(client:)
       @client = client
       @websocket = nil
+      @on_message = nil
     end
 
-    def connect(model: "gpt-4o-realtime-preview-2024-10-01", &block)
+    def on_message(&block)
+      @on_message = block
+    end
+
+    def connect(model: "gpt-4o-realtime-preview-2024-10-01")
       uri = "#{File.join(@client.websocket_uri_base, @client.api_version, 'realtime')}?model=#{model}"
       headers = {
         "Authorization" => "Bearer #{@client.access_token}",
@@ -16,7 +21,7 @@ module OpenAI
       }
       EM.run do
         @websocket = Faye::WebSocket::Client.new(uri, nil, headers: headers)
-        @websocket.on :message, &block
+        @websocket.on :message, @on_message
       end
     end
 
--- a/lib/openai/real_time.rb
+++ b/lib/openai/real_time.rb
@@ -3,6 +3,8 @@ require "eventmachine"
 
 module OpenAI
   class RealTime
+    include HTTPHeaders
+
     def initialize(client:)
       @client = client
       @websocket = nil
@@ -15,12 +17,9 @@ module OpenAI
 
     def connect(model: "gpt-4o-realtime-preview-2024-10-01")
       uri = "#{File.join(@client.websocket_uri_base, @client.api_version, 'realtime')}?model=#{model}"
-      headers = {
-        "Authorization" => "Bearer #{@client.access_token}",
-        "OpenAI-Beta" => "realtime=v1"
-      }
+
       EM.run do
-        @websocket = Faye::WebSocket::Client.new(uri, nil, headers: headers)
+        @websocket = Faye::WebSocket::Client.new(uri, nil, headers: openai_realtime_headers)
         @websocket.on :message, @on_message
       end
     end
--- a/lib/openai/real_time.rb
+++ b/lib/openai/real_time.rb
@@ -8,14 +8,14 @@ module OpenAI
       @websocket = nil
     end
 
-    def connect(&block)
-      url = "wss://api.openai.com/v1/realtime?model=gpt-4o-realtime-preview-2024-10-01"
+    def connect(model: "gpt-4o-realtime-preview-2024-10-01", &block)
+      uri = "#{File.join(@client.websocket_uri_base, @client.api_version, 'realtime')}?model=#{model}"
       headers = {
         "Authorization" => "Bearer #{@client.access_token}",
         "OpenAI-Beta" => "realtime=v1"
       }
       EM.run do
-        @websocket = Faye::WebSocket::Client.new(url, nil, headers: headers)
+        @websocket = Faye::WebSocket::Client.new(uri, nil, headers: headers)
         @websocket.on :message, &block
       end
     end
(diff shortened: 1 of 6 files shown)


lib/openai/real_time.rb
```diff
@@ -0,0 +1,31 @@
...
+      @websocket = nil
+      @on_message = nil
+    end
+
+    def on_message(&block)
+      @on_message = block
+    end
+
+    def connect(model: "gpt-4o-realtime-preview-2024-10-01")
+      uri = "#{File.join(@client.websocket_uri_base, @client.api_version, 'realtime')}?model=#{model}"
```
franpb14, line 19: probably this uri shouldn't be here but I was not sure where to put it

drnic:
Out of scope to add a sample sinatra app into the repo with some stimulusjs that demos the client-side setup of the websocket/connection to feed/receive messages to the backend?

lib/openai/real_time.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      EM.run do
+        @websocket = Faye::WebSocket::Client.new(uri, nil, headers: openai_realtime_headers)
+        @websocket.on :message, @on_message
+      end
```
franpb14, line 24: Perhaps we should replace [Eventmachine](https://github.com/eventmachine/eventmachine) since it last release was 6 yeas ago. I think we could use [async](https://github.com/socketry/async). I have tried and it works fine with something like this:
```
    Async do
      endpoint = Async::HTTP::Endpoint.parse(uri, alpn_protocols: Async::HTTP::Protocol::HTTP11.names)
      Async::WebSocket::Client.connect(endpoint, headers: @headers) do |connection|
        @websocket = connection

        while (message = connection.read)
          @on_message
        end
      end
    end
```

franpb14:
@drnic Sorry the delay in the answer, I think it'd be a good thing to have, for me it was the hardest part. If we finally merge this I  could do that app easily.

ngelx:
What is the status of this PR? looks likes just what it is need to ease RealTime implementation.

alexrudall:
Thanks @ngelx - Realtime is my top priority for v8.1 - would you be able to test this PR and see how useful you find it?

franpb14:
@drnic @ngelx since I did this PR, openAI has introduced the possibility of doing it with webRTC https://platform.openai.com/docs/guides/realtime-webrtc and it works pretty well, by using it you don't need to set up Faye or other dependency, I'm not sure if this PR makes sense and we should only include the endpoint to get the ephemeral key or if we should have both possibilities. What do you think?

ngelx:
> @alexrudall @ngelx since I did this PR, openAI has introduced the possibility of doing it with webRTC https://platform.openai.com/docs/guides/realtime-webrtc and it works pretty well, by using it you don't need to set up Faye or other dependency, I'm not sure if this PR makes sense and we should only include the endpoint to get the ephemeral key or if we should have both possibilities. What do you think?

Why not both?

As you mentioned, the WebRTC implementation only requires exposing the ephemeral key endpoint, leaving the rest to the client. Since it's part of the API, it makes sense to support it.

On the other hand, the WebSocket implementation requires more backend work but simplifies the client-side integration. It's also officially beta supported by other SDKs (e.g., [openai-python realtime api](https://github.com/openai/openai-python/blob/main/api.md#realtime)).

To sum up, @franpb14  raises a valid point by suggesting WebRTC. Maybe @alexrudall  already has plans to support both options?

alexrudall:
> > @alexrudall @ngelx since I did this PR, openAI has introduced the possibility of doing it with webRTC https://platform.openai.com/docs/guides/realtime-webrtc and it works pretty well, by using it you don't need to set up Faye or other dependency, I'm not sure if this PR makes sense and we should only include the endpoint to get the ephemeral key or if we should have both possibilities. What do you think?
> 
> Why not both?
> 
> As you mentioned, the WebRTC implementation only requires exposing the ephemeral key endpoint, leaving the rest to the client. Since it's part of the API, it makes sense to support it.
> 
> On the other hand, the WebSocket implementation requires more backend work but simplifies the client-side integration. It's also officially beta supported by other SDKs (e.g., [openai-python realtime api](https://github.com/openai/openai-python/blob/main/api.md#realtime)).
> 
> To sum up, @franpb14 raises a valid point by suggesting WebRTC. Maybe @alexrudall already has plans to support both options?

Agree I think. Although I need to understand it better. If there's one way that makes it easier and simpler for the user I prefer to support that and only that, even if it's more work in the gem, but in this case maybe both make sense. Have you seen [this thread](https://github.com/alexrudall/ruby-openai/issues/524)?

ngelx:
Sorry for the delay on the reply. I did try this fork and did work, but in particular for my application, was too laggy. So i went for the WebRTC implementation.

I did a separate PR #582 but following the structure of this one so they can be easily merged. The code is quite simple. The heavy part is in the web client, but i guess that is what WebRTC is all about.

alexrudall:
Just released @ngelx 's WebRTC PR, #582, in [v8.2.0](https://github.com/alexrudall/ruby-openai/releases). Still open to the websockets side also, if there's demand. This PR would need to rebased and rewritten a bit to work with the existing Realtime class now.

PR #550, opened 2024-11-28
Async fix

atesgoral:
## All Submissions:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 # Include gem dependencies from ruby-openai.gemspec
 gemspec
 
+gem "async-http-faraday", "~> 0.19.0"
 gem "byebug", "~> 11.1.3"
 gem "dotenv", "~> 2.8.1"
 gem "rake", "~> 13.2"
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 # Include gem dependencies from ruby-openai.gemspec
 gemspec
 
+gem "async-http-faraday", "~> 0.19.0"
```
atesgoral, line 6: Won't be part of gem deps, but will solely be used in tests.

lib/openai/client.rb
```diff
@@ -12,6 +12,7 @@ class Client
       uri_base
       request_timeout
       extra_headers
+      async
```
atesgoral, line 15: New config option

lib/openai/http.rb
```diff
@@ -78,6 +78,7 @@ def conn(multipart: false)
         f.use MiddlewareErrors if @log_errors
         f.response :raise_error
         f.response :json
+        f.adapter :async_http if @async
```
atesgoral, line 81: This will require users to include `async-http-faraday` in their app themselves.

PR #553, opened 2024-12-02
ignore ruby-version, fix chat and completions api for azure

gbleydon:
## All Submissions:

* [x ] Have you followed the guidelines in our [Contributing document](../blob/main/CONTRIBUTING.md)?
* [x ] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x ] Have you added an explanation of what your changes do and why you'd like us to include them?

Fixed the issue when using azure and trying to use either /chat or /completions api. Azure requires different base uri for /completions api. Extracted changes out to build_azure_uri and added parameters: nil to signature of uri method and adjusted calls in http.rb to add the parameters argument when available. Need access to model parameter to build the azure base_uri for /chat or /chat/completions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 ### Ruby ###
+.ruby-version
 *.gem
 *.rbc
 /.config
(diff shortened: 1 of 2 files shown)


gbleydon:
Had a bug in /chat that I missed will resubmit a clean fix

PR #559, opened 2025-01-15
Add pricing calculation for OpenAI API responses (chat, embeddings and completions)

F-3r:
- Add new OpenAI::Pricing class to calculate costs based on token usage
- Integrate automatic cost calculation into API response handling
- Include detailed cost breakdown for:
  - Input tokens (regular, cached, audio)
  - Output tokens (regular, reasoning, audio, predictions)

The pricing data supports all current OpenAI models including GPT-4, embeddings, and audio-enabled variants.
Costs are automatically merged into API responses under a new 'cost' key that follows the structure of the "usage" key

Bonus:
- a script to fetch latest model costs from OpenAI website

## All Submissions:

### DIFF
--- a/lib/openai/pricing.rb
+++ b/lib/openai/pricing.rb
@@ -1,34 +1,19 @@
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
 module OpenAI
   class Pricing
-    COSTS = {
-      "gpt-4o" => {input: 2.5, cached_input: 1.25, output: 10.0, audio_input: nil, audio_output: nil},
-      "gpt-4o-2024-11-20" => {input: 2.5, cached_input: 1.25, output: 10.0, audio_input: nil, audio_output: nil},
-      "gpt-4o-2024-08-06" => {input: 2.5, cached_input: 1.25, output: 10.0, audio_input: nil, audio_output: nil},
-      "gpt-4o-audio-preview" => {input: 2.5, cached_input: nil, output: 10.0, audio_input: 40.0, audio_output: 80.0},
-      "gpt-4o-audio-preview-2024-12-17" => {input: 2.5, cached_input: nil, output: 10.0, audio_input: 40.0, audio_output: 80.0},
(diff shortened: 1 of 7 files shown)


lib/openai/pricing.rb
```diff
@@ -0,0 +1,104 @@
+module OpenAI
+  class Pricing
+    COSTS = {
```
MadBomber, line 3: How about a class variable that gets loaded with this default data but is updatable from either a JSON file or the website.
```ruby
DEFAULT_COSTS = {...}
@@costs ||= DEFAULT_COSTS
def self.costs = @@costs
def costs = self.class.costs

def self.update_costs(json_file: nil)
  If json_file.nil?
    # get it from the website
  elsif json_file is a bad file
    raise an error
  else
    parse that json file into @@costs
  end
end
```
I asked my robot for an example for this pattern and it gave me this ...

```ruby
class CostManager
  DEFAULT_COSTS = {
    apple:  1.00,
    banana: 0.50,
    orange: 0.75
  }

  @@costs ||= DEFAULT_COSTS

  # Class method to access costs
  def self.costs = @@costs

  # Instance method to access costs
  def costs = self.class.costs

  # Updates costs from a JSON file
  # @param json_file [String] Path to the JSON file
  def self.update_costs(json_file: nil)
    if json_file.nil?
      @@costs = self.class.get_it_from_the_website
    elsif !File.exist?(json_file)
      raise ArgumentError, "JSON file does not exist"
    else
      begin
        # Parse JSON file and update costs
        new_costs = JSON.parse(File.read(json_file),
                               symbolize_names: true)
        @@costs   = new_costs
      rescue JSON::ParserError
        raise ArgumentError, "Invalid JSON format in file"
      end
    end
  end
end
```

MadBomber (changes requested):

alexrudall:
Thank you for this PR @F-3r ! In quite amazing timing, OpenAI just released a Usage API 😬[https://platform.openai.com/docs/api-reference/usage/costs](https://platform.openai.com/docs/api-reference/usage/costs)

MadBomber:
Yep, I like it coming from the JSON file much better.  That way as an app
dev using the gem, if I have a different set of prices all I have to do is
use my own JSON file.

Dewayne
  o-*

Issue #562, opened 2025-01-29
Retrieving contents of CSV file errors on JSON parsing


nuclearghost:
**Describe the bug**
The OpenAI Assistant has generated a CSV file. When I attempt to retrieve its contents with `openai_client.files.content(id: 'file_id')` it generates the error `unexpected character (after [0]) at line 1, column 2 [parse.c:744]` which appears to be the JSON parser trying to parse CSV content.

**To Reproduce**
```
openai_client.files.retrieve id: "file-NexuEUjWpoehzmkQ5zer7W"
{"object"=>"file",
 "id"=>"file-NexuEUjWpoehzmkQ5zer7W",
 "purpose"=>"assistants_output",
 "filename"=>"/mnt/data/normalized_file.csv",
 "bytes"=>1630,
 "created_at"=>1738176534,
 "status"=>"processed",
 "status_details"=>nil}
openai_client.files.content id: "file-NexuEUjWpoehzmkQ5zer7W"
: unexpected character (after [0]) at line 1, column 2 [parse.c:744]
```

**Expected behavior**
The CSV file contents are able to be read without error.

**Desktop (please complete the following information):**
 - OS: macOS 15.2
 - Ruby version: 3.2.2
 - Gem version: 7.3.1

james-jang:
Same issue. Appears to be coming from 
https://github.com/alexrudall/ruby-openai/blob/main/lib/openai/http.rb#L10 
because `content` simply calls `.get` and assumes that everything returning is a jsonl https://github.com/alexrudall/ruby-openai/blob/c87e36ec122b15049da10e9b0438fc8b463133a9/lib/openai/files.rb#L37
Will need to make a direct Faraday call for now..

alexrudall:
I believe https://github.com/alexrudall/ruby-openai/pull/566 resolves this issue - if you are able to could you give it a test @nuclearghost , @james-jang ? Thank you!

alexrudall:
Fixed in [ruby-openai v8](https://rubygems.org/gems/ruby-openai)

PR #544, opened 2024-10-29
Fix - File content

gregszero:
This pull request includes changes to the `get` method in the `HTTP` module within the `lib/openai/http.rb` file. The main purpose of these changes is to improve error handling when parsing JSON responses to address issue #421

Error handling improvements:

* [`lib/openai/http.rb`](diffhunk://#diff-b944e7fce4884db7a0639a637f10d5a081ca198b280570d874865ee773a6bc09L10-R18): Modified the `get` method to handle `JSON::ParserError` exceptions. If parsing the response body as JSON fails, the raw response body will be returned instead.

## All Submissions:

* [x] Have you followed the guidelines in our [Contributing document](../blob/main/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

> there’s another PR addressing this issue, but since I found a solution that works for me, I believe the community could also benefit from this change

PR #540, opened 2024-10-21
fix: vtt transcribe not working issue

qoosuperman:
## Why
relative issue: https://github.com/alexrudall/ruby-openai/issues/277

## Description
when using transcribe api with vtt response format, output would be the content of VTT file instead of string in JSON format

## All Submissions:

* [x] Have you followed the guidelines in our [Contributing document](../blob/main/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x] Have you added an explanation of what your changes do and why you'd like us to include them?

Issue #416, opened 2023-12-26
When json_post is executed, is JSON.parse not executed on the response?


tonegawa07:
I noticed that in the recent commit (dc4d1c6b8d0d52ae5897e8203912b9b047ac981f) on https://github.com/alexrudall/ruby-openai, the processing for JSON.parse within the json_post method seems to have been removed. Could you please clarify the reason behind this change?

https://github.com/alexrudall/ruby-openai/commit/dc4d1c6b8d0d52ae5897e8203912b9b047ac981f

After creating a stub with WebMock as shown below, I encountered an error during the execution of dig('choices', 0, 'message', 'content') due to the absence of JSON.parse being executed.

```rb
response = {
  "model" => "gpt-4-0613",
  "choices" => [
    {
      "message" => {
        "role" => "assistant",
        "content" => "Response from the API",
      },
      "finish_reason" => "stop",
      "index" => 0
    }
  ]
}

stub_request(:post, "https://api.openai.com/v1/chat/completions")
  .to_return(status: 200, body: response.to_json, headers: {})
```

```
ERROR -- : undefined method `dig' for "{\"model\":\"gpt-4-0613\",\"choices\":[{\"message\":{\"role\":\"assistant\",\"content\":\"Response from the API\"},\"finish_reason\":\"stop\",\"index\":0}]}":String

response.dig('choices', 0, 'message', 'content')
                ^^^^ (NoMethodError)
```

It seems like the removal of JSON.parse from json_post might be the cause of this issue. What are your thoughts on this?

jstewmc:
I found this issue when I was searching for a solution to the same problem. It's been a minute; and, I don't know if y'all are still having it; but, I found the solution. 

Set the request's content type:

```ruby
stub_request(:post, "https://api.openai.com/v1/chat/completions")
  .to_return(status: 200, body: response.to_json, headers: { "Content-Type" => "application/json" })
```

alexrudall:
I believe #566 resolves this issue - if you are able to could you give it a test @gjtorikian, @jstewmc, @tonegawa07 ? Thank you!

alexrudall:
Fixed in [ruby-openai v8](https://rubygems.org/gems/ruby-openai)

PR #566, opened 2025-02-14
Fixes for various JSON/non-JSON content issues

alexrudall:
[BREAKING CHANGE]

- Try to parse all responses as JSON
- If parsing JSON fails, return the original response as-is
- Reproduction and fix for [https://github.com/alexrudall/ruby-openai/pull/544](https://github.com/alexrudall/ruby-openai/pull/544)

## To Test

Update gemfile:

```
gem 'ruby-openai', git: 'git://github.com/alexrudall/ruby-openai.git', branch: 'fix/non-json-content'
```

Run `bundle install`

## Related Issues (?)
#277 
#314
#416
#421 
#562 

## Related PRs (?)
#474
#504 
#540 
#544

## All Submissions:

### DIFF
--- a/spec/openai/client/files_spec.rb
+++ b/spec/openai/client/files_spec.rb
@@ -128,5 +128,32 @@ RSpec.describe OpenAI::Client do
         end
       end
     end
+
+    describe "#fetch_image" do
+      let(:cassette) { "files fetch_image" }
+      let(:upload_cassette) { "#{cassette} upload" }
+      let(:filename) { "image.png" }
+      let(:file) { File.join(RSPEC_ROOT, "fixtures/files", filename) }
+      let(:upload_purpose) { "vision" }
+
+      def poll_until_processed(max_attempts: 10)
+        VCR.use_cassette("#{cassette}_poll", record: :new_episodes) do
+          max_attempts.times do |attempt|
+            retrieved = OpenAI::Client.new.files.retrieve(id: upload_id)
+            return retrieved if retrieved["status"] == "processed"
+            raise "File not processed after #{max_attempts} attempts" if attempt == max_attempts - 1
+          end
+        end
+      end
+
+      it "succeeds in uploading and retrieving an image" do
+        VCR.use_cassette(cassette) do
+          poll_until_processed
+          response = OpenAI::Client.new.files.content(id: upload_id)
+          expect(response).to be_a(String)
+          expect(response.size).to be > 0
+        end
+      end
+    end
   end
 end
--- a/spec/openai/client/files_spec.rb
+++ b/spec/openai/client/files_spec.rb
@@ -132,24 +132,32 @@ RSpec.describe OpenAI::Client do
     describe "#fetch_image" do
       let(:cassette) { "files fetch_image" }
       let(:upload_cassette) { "#{cassette} upload" }
+      let(:retrieve_cassette) { "#{cassette} retrieve" }
       let(:filename) { "image.png" }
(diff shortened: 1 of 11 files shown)


spec/openai/client/files_spec.rb
```diff
@@ -128,5 +128,32 @@
         end
       end
     end
+
+    describe "#fetch_image" do
+      let(:cassette) { "files fetch_image" }
+      let(:upload_cassette) { "#{cassette} upload" }
+      let(:filename) { "image.png" }
+      let(:file) { File.join(RSPEC_ROOT, "fixtures/files", filename) }
+      let(:upload_purpose) { "vision" }
```
ndemianc, line 137: Seems like this is not used anywhere.

ndemianc (approved):

spec/openai/client/files_spec.rb
alexrudall: It's hard to read, but it's used in setup via upload_id (line 142) -> upload, and then upload_cassette is used there too. It's so each spec has its own setup cassettes.

alexrudall: Refactored spec to be more consistent with the one above.

spec/openai/client/files_spec.rb
```diff
@@ -132,24 +132,32 @@
...
+        until retrieved["status"] == "processed"
+          raise "File not processed after 10 tries" if tries > 10
 
-      def poll_until_processed(max_attempts: 10)
-        VCR.use_cassette("#{cassette}_poll", record: :new_episodes) do
-          max_attempts.times do |attempt|
-            retrieved = OpenAI::Client.new.files.retrieve(id: upload_id)
-            return retrieved if retrieved["status"] == "processed"
-            raise "File not processed after #{max_attempts} attempts" if attempt == max_attempts - 1
+          sleep(1)
```
ndemianc, line 151: Never liked the `sleep` statements, but it seems like it serves it's purpose, and this is still in the tests so can be refactored later (for example with https://github.com/ruby/timeout)

spec/openai/client/files_spec.rb
ndemianc: Got it! Thanks!

ndemianc (approved):

spec/openai/client/files_spec.rb
alexrudall: agree. plus behind VCR so it doesn't get hit except when re-recording cassettes

alexrudall: Thank you for review!

jprosevear:
Is this just about the files endpoint or is there others that don't return JSON?

alexrudall:
@jprosevear Audio endpoints are another one

jprosevear:
> @jprosevear Audio endpoints are another one

My thought was to pass a named param to select the parsing option and default to json.  Then change the default for files and audio to limit the breaking damage.  Not sure if that's the right path though

alexrudall:
> > @jprosevear Audio endpoints are another one
> 
> My thought was to pass a named param to select the parsing option and default to json.  Then change the default for files and audio to limit the breaking damage.  Not sure if that's the right path though 

Yep, saw that in one of the PRs. I'd prefer just just always check if Strings can be JSON and always return raw if there are any errors - that way it should work for future endpoints, unforseen edge cases, etc. not sure of downsides to this approach

alexrudall:
Released in [ruby-openai v8](https://rubygems.org/gems/ruby-openai)

PR #589, opened 2025-05-01
Refactor stream handling to expose event labels

ingemar:
### Description:

This PR refactors how streaming responses are handled to [make stream event labels](https://platform.openai.com/docs/api-reference/assistants-streaming/events) accessible to the user-defined stream handler.

### Why This Change?

This update enables access to the event labels in streamed OpenAI API responses. These labels provide important context (e.g., `response.created`, `response.completed`) previously unavailable to user-defined stream handlers.

Exposing this metadata makes building more responsive and structured streaming experiences easier, while keeping the API flexible and backward compatible.

### Summary of Changes

- Moved the code from `OpenAI::HTTP#to_json_stream` into a new class named `OpenAI::Stream`
- `OpenAI::Stream` maintains backward compatibility by inspecting the arity of the proc and calling it with one or two arguments accordingly
- Replaced all calls to the `#to_json_stream` method with the new `OpenAI::Stream` interface
- Updated all examples and specs to replace unused `_bytesize` params with `_event` (since, as far as I understand, the bytesize was never actually passed)

### Example Before vs. After

**Before:**
```ruby
  puts chunk["delta"]
end
```

**After:**
```ruby
stream: proc do |chunk, event|
  puts "[#{event}] #{chunk["delta"]}"
end
```

### Compatibility

- Fully backward compatible: if a stream proc only takes one argument, it will behave as before
- All tests passing

---

## All Submissions:

```

### CODE REVIEW

File: lib/openai/stream.rb [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+
+      parser.feed(chunk) do |event, data|
+        next if data == DONE
+
+        args = [JSON.parse(data), event].first(user_proc_arity)
+        user_proc.call(*args)
+      end
+    end
+
+    def to_proc
```
ingemar, line 31: We need to convert to a `Proc` because [Faraday enforces](https://github.com/lostisland/faraday/blob/4018769a30a1c16b48289e7120892ef92102417f/lib/faraday/options/request_options.rb#L19) the stream handler to be that.

alexrudall:
Released in [v8.2.0](https://github.com/alexrudall/ruby-openai/releases/tag/v8.2.0) - thank you for this excellent PR @ingemar !

ingemar:
@alexrudall Thank you! <3

PR #618, opened 2025-08-17
Fix `ArgumentError` in `Stream#call` with Faraday 2.1.0

### LINKED ISSUE #617: Stream broken in v8.2 with Faraday/NetHTTP adapter
**Describe the bug**
Getting an error inside Faraday's `on_data` callback after upgrading from 8.1 to 8.2.

```
#<ArgumentError: wrong number of arguments (given 2, expected 3)>
```

Looks like the problem is that the gem is setting Faraday's `on_data` callback with invalid arguments, these are the last 2 lines in the call stack:

```
"/Users/fabiogomes/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/gems/ruby-openai-8.2.0/lib/openai/stream.rb:20:in 'call'",
 "/Users/fabiogomes/.asdf/installs/ruby/3.4.4/lib/ruby/gems/3.4.0/gems/faraday-net_http-2.1.0/lib/faraday/adapter/net_http.rb:112:in 'block in Faraday::Adapter::NetHttp#perform_request'",
````

Here's the Faraday code that is trying to call the function:

```ruby
env[:request].on_data.call(chunk, size)
```

And the function definition:

```ruby
def call(chunk, _bytes, env)
      handle_http_error(chunk: chunk, env: env) if env && env.status != 200
```

**To Reproduce**
Use stream with Faraday and NetHTTP with version 8.2

daikimiura:
Fixes #617 

## All Submissions:

## Problem

After upgrading from ruby-openai 8.1 to 8.2, users encountered an `ArgumentError` when using streaming with Faraday 2.1.0:

  `ArgumentError: wrong number of arguments (given 2, expected 3)`

  The error occurs because Faraday 2.1.0 calls the `on_data` callback with only 2 arguments (`chunk`, `size`), but the new `Stream#call` method expects 3 arguments (`chunk`, `_bytes`, `env`).

## Root Cause

  In version 8.2.0, streaming functionality was refactored from an inline Proc in `HTTP#to_json_stream` to a dedicated`Stream` class (PR #589). This introduced a breaking change:

  **Before 8.2.0 (worked fine):**
  - Used a Proc: `proc do |chunk, _bytes, env|`
  - Ruby Procs are flexible with argument counts - when called with
  fewer arguments, missing parameters become `nil`
  - Faraday calling with 2 arguments worked without errors

  **After 8.2.0 (breaks):**
  - Uses a method: `def call(chunk, _bytes, env)`
  - Ruby methods strictly enforce argument counts
  - Faraday calling with 2 arguments raises `ArgumentError`

  ## Faraday's Behavior

  In `faraday-net_http` v2.1.0, the callback is invoked with only 2 parameters:
  ```ruby
  # https://github.com/lostisland/faraday-net_http/blob/v2.1.0/lib/faraday/adapter/net_http.rb#L112
  env[:request].on_data.call(chunk, size)

  ## Solution

  Make the env parameter optional with a default value of nil:

  ```def call(chunk, _bytes, env = nil)```

  This maintains backward compatibility with different Faraday versions while preserving the same functionality.

### DIFF
--- a/lib/openai/stream.rb
+++ b/lib/openai/stream.rb
@@ -17,7 +17,7 @@ module OpenAI
         end
     end
 
-    def call(chunk, _bytes, env)
+    def call(chunk, _bytes, env = nil)
       handle_http_error(chunk: chunk, env: env) if env && env.status != 200
 
       parser.feed(chunk) do |event, data|
--- a/spec/openai/client/stream_spec.rb
+++ b/spec/openai/client/stream_spec.rb
@@ -106,6 +106,26 @@ RSpec.describe OpenAI::Stream do
           CHUNK
         end
       end
+
+      context "when called with only 2 arguments (like Faraday 2.1.0)" do
+        it "handles the call without env parameter" do
+          expect(user_proc).to receive(:call)
+            .with(
+              JSON.parse('{"foo": "bar"}'),
+              "event.test"
+            )
+
+          # Faraday 2.1.0 calls with only (chunk, size), not (chunk, size, env)
+          expect do
+            stream.call(<<~CHUNK, bytes)
+              event: event.test
+              data: { "foo": "bar" }
+
+              #
+            CHUNK
+          end.not_to raise_error
+        end
+      end
     end
   end
 


JadJabbour:
when will these be merged?

alexrudall:
Thanks for this excellent PR @daikimiura - released in [8.2.1](https://github.com/alexrudall/ruby-openai/releases/tag/v8.2.1)

PR #626, opened 2025-09-23
adding back compatibility layer

### LINKED ISSUE #625: Incompatible with official OpenAI library
It seems this library came out before OpenAI released their [official Ruby library](https://github.com/openai/openai-ruby), but that's been out for awhile now. Unfortunately these two libraries directly conflict with one another.

If you add both "openai" and "ruby-openai" to your Gemfile, whichever one gets loaded first will predominate. Both libraries use `require "openai"` as their convention making them incompatible. I would have expected that this one would use `require "ruby-openai"` which would have enabled some possibility of using both. And that's the desired solution here: change the require path for this to match its own gem name.

The trouble is that I'm in a codebase which is using this library and I'm trying to cut over to the official client library instead. However it's impossible to do any sort of phased rollout because of this compatibility issue. As things stand, I have to do a hard cutover or nothing.

soapergem:
Fixes #625. Adds back the compatibility layer that allows this library to be included via `require "ruby/openai"` so that it can cohabitate with OpenAI's official client library for Ruby. Please let me know if I missed anything here.

## All Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,14 @@ and require with:
 require "openai"
 ```
 
+## Compatibility with offical OpenAI library
+
+OpenAI now publishes their own [Ruby client library](https://github.com/openai/openai-ruby), which uses the same namespace as this one. If you are working with a codebase that uses both, you can require this one with this alternate path to avoid collisions:
+
+```ruby
+require "ruby/openai"
+```
+
 ## How to use
 
 - Get your API key from [https://platform.openai.com/account/api-keys](https://platform.openai.com/account/api-keys)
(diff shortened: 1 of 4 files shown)


alexrudall:
Thanks @soapergem - ~looking into this, I believe using `require "ruby/openai"` should just work without this PR required. Did you already try that, if it doesn't work can I ask what error you get?~

Edit: Thinking more about this, I guess the clash will still occur because even if you require with `require ruby/openai`, `OpenAI::Client.new` will still work for both gems. So we would need to figure out a way to force ruby-openai to only use `Ruby::OpenAI` in compatibility mode, if that's even possible. I wonder if reverting back to v3.2 would work for you, depending on which features you're using?

